PR #1511, opened 2020-09-29
Custom Draft Cover

robertu7:
* Remove fallback and reset cover in `putDraft`
* Prevent deleting cover asset in `deleteUnusedAssets` 

https://github.com/thematters/matters-web/issues/1527


devformatters (approved):
Go. 🦿

PR #1514, opened 2020-09-30
ci(migration): comment unused console output, but keep for reading

devformatters:
Comment unused console output, but keep for who wants to read codes. 🧑🏻‍💻


robertu7 (approved):
mute them all.

jeremyOk (approved):

PR #1533, opened 2020-10-12
Update article query resolver due to draft and article data restructure

devformatters:
As title.

### DIFF
--- a/src/mutations/article/updateTagSetting.ts
+++ b/src/mutations/article/updateTagSetting.ts
@@ -259,7 +259,7 @@ const resolver: MutationToUpdateTagSettingResolver = async (
 
       // send notice
       if (tag.owner) {
-        notificationService.trigger({
+        await notificationService.trigger({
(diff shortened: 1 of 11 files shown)


robertu7 (approved):
Nice hard work!

I thought it's a safe method to resolve `content`, `dataHash` and etc from draft, but what if we merge draft and article at root level (`rootArticle.ts` or knex), will it more simple and consistent? Especially when we support resolving version article.

devformatters:
> Nice hard work!
> 
> I thought it's a safe method to resolve `content`, `dataHash` and etc from draft, but what if we merge draft and article at root level (`rootArticle.ts` or knex), will it more simple and consistent? Especially when we support resolving version article.

It sounds fine to me. My way is to avoid touching root and break logics into resolvers. In this way, we can revise and remove single resolver without affecting the root object.

devformatters:
I'll merge it first, and then we can discuss in meeting.

PR #1534, opened 2020-10-13
feat(search api): add `record` boolean to input

guoliu:
Add `record` boolean to search input to determine if search key needs to be stored.


devformatters (approved):
Go. 🚖

PR #1538, opened 2020-10-13
fix: a/b grouping logic, update hottest group b view param

Carolusian:
- Take `admin` user into account when grouping users
- Update decay factor to 0.8 for hottest articles

### DIFF
--- a/src/common/utils/getViewer.ts
+++ b/src/common/utils/getViewer.ts
@@ -32,9 +32,11 @@ export const authModes = [
 export const getUserGroup = ({
   id,
   ip,
+  role,
 }: {
   id?: string | null
   ip?: string
+  role?: string | null
 }) => {
   let num = 0
   try {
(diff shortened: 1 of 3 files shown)


src/common/utils/getViewer.ts
```diff
@@ -47,6 +49,10 @@ export const getUserGroup = ({
   } catch (error) {
     logger.error(error)
   }
+
+  if (role === USER_ROLE.admin) {
+    return 'b'
```
robertu7, line 54: any related discussion on this change?

robertu7 (approved):

src/common/utils/getViewer.ts
Carolusian: Yes, @jeremyOk proposed this at wrote the logic in the SQL which is used in migrations: https://github.com/thematters/matters-server/pull/1538/files#diff-b726e1a846dbc8b25876ac766b0ceaa41f4983702650e221ea1a80e84a876a11R23

devformatters (approved):
Go. 🚟

guoliu (approved):

Carolusian:
This PR will be closed. As the cookie forwarding mechanism updated

PR #1539, opened 2020-10-13
feat(resolver): revise api to return tags that user own and maintain

devformatters:
As title.

### DIFF
--- a/src/connectors/tagService.ts
+++ b/src/connectors/tagService.ts
@@ -110,6 +110,18 @@ export class TagService extends BaseService {
       .orderBy('id', 'desc')
 
   /**
+   * Find tags by a given maintainer id (user).
+   *
+   */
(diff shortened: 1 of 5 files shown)


src/connectors/tagService.ts
```diff
@@ -109,6 +109,18 @@ export class TagService extends BaseService {
...
+  /**
+   * Find tags by a given maintainer id (user).
+   *
+   */
+  findByMaintainer = async (userId: string) =>
+    this.knex
+      .select()
+      .from(this.table)
+      .where({ owner: userId })
+      .orWhere(this.knex.raw(`editors @> ARRAY['${userId}']`))
```
guoliu, line 121: nice

guoliu (approved):
🍍

PR #1548, opened 2020-10-19
fix(asset): permission checking inside resolver instead of using `@auth`

robertu7:
Check permission inside resolver instead of `@auth(mode: "${AUTH_MODE.oauth}")` since `@auth` now only supports scope starting with `viewer`, which means the root type of the resolver is `User` and `Article.assets` [can't match](https://github.com/thematters/matters-server/blob/develop/src/types/directives/auth.ts#L22).

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -112,7 +112,7 @@ type Article implements Node {
   """Article cover's link."""
   cover: URL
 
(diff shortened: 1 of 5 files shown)


devformatters (approved):
It's the simplest way to do ti. Go. 🧼

PR #1549, opened 2020-10-19
feat(search): shorten frequent search interval

robertu7:
* Use `findUnpublishedByAuthor` as resolver of `User.drafts`;
* Shorten frequent search interval


devformatters (approved):
Go. 🔭

PR #1553, opened 2020-10-20
Getter & Setter of User Group

robertu7:
* Set `__user_group` cookie with disabled httpOnly and under SLD
* Caching public query based on user group


Carolusian (approved):

guoliu (approved):
🌴

PR #1557, opened 2020-10-21
feat(api): add api for listing article's drafts and create queue for publish

devformatters:
As title.


src/connectors/queue/revision.ts
```diff
@@ -0,0 +1,310 @@
...
+          entityTypeId,
+          entityId: preDraft.id,
+        })
+      ).map((assetMap) => assetMap.id)
+      await this.systemService.swapAssetMapEntity(
+        assetMapIds,
+        entityTypeId,
+        draft.id
+      )
+      job.progress(60)
```
devformatters, line 179: @guoliu I'm not sure we should transfer previous draft's assets to new one. The only advantage is if we allow user to drop previous drafts, then assets can be kept. Any opinion?

cc @robertu7

guoliu: @devformatters I think we shouldn't allow newer draft (or article version) to inherit assets from older ones, as long as the newer draft already have the assets it needs. So that when we do need to delete drafts, we can cascade delete assets that are not used by other drafts.

So I'm not sure about this part here: if in other places we already assigned the needed assets to the new draft, we don't need this part, but if we don't then we still need to keep it.

guoliu (approved):
Looks good 🍃

src/connectors/queue/revision.ts
devformatters: Based on last discussion, current revision feature won't allow users to change embed assets. So, the simplest way is to copy asset maps. I've changed codes.

guoliu (approved):
Looks good!

PR #1562, opened 2020-10-23
fix(cache): fix article cache key

robertu7:


devformatters (approved):
Very nice improvement. 🖖🏻

PR #1574, opened 2020-11-03
fix(slug): fix getting wrong slug when recording read and super like

devformatters:
Since `slug` field will be removed from `Article` table, fix it using slugify instead of getting from `Article` table.


guoliu (approved):
🌾

PR #1584, opened 2020-11-06
fix(article read): count visitor read as new read when ip changes

guoliu:
Count visitor read as new read when IP id different than record.


devformatters (approved):
Go. 👣

Issue #1596, opened 2020-11-12
Multiple articles map to one draft


guoliu:
There are at least two instances where multiple articles are map to the same draft, result in identical articles:
* https://matters.news/@clarinet 
* https://matters.news/@conanxin

devformatters:
Looks like this problem exists before refactoring draft and article data structure.

```
SELECT
	*
FROM (
	SELECT
		count(id) AS count,
		draft_id
	FROM
		article
	GROUP BY
		draft_id) AS source
	JOIN article ON article.draft_id = source.draft_id
WHERE
	count > 1
ORDER BY
	source.draft_id ASC
```

PR #1609, opened 2020-11-19
Fix multiple articles map to same draft

robertu7:
**Changes of publication queue** 
* mark `publishState` as `error` on publication error;
  * ignore errors caused by steps like tags handling since they aren't the core and can be added back later;
* break down `articleService.public` into `articleService.publishToIPFS` and `articleService.createArticle`;
* remove `publishPendingDrafts` queue;

**Changes of revision queue** 
* mark `publishState` as `error` and trigger notification on revision error;
  * ignore errors caused by steps like asset swapping since they aren't the core and can be added back later;
* remove `publishPendingRevisionDrafts ` queue;


guoliu (approved):
LGTM 🎄

We also need to do a migration to remove the duplicated articles. But we can do that in a separate PR.

devformatters (approved):
Looks good. ⛩

robertu7:
> LGTM 🎄
> 
> We also need to do a migration to remove the duplicated articles. But we can do that in a separate PR.

As @devformatters [mentioned](https://github.com/thematters/matters-server/issues/1596#issuecomment-730074639), there are ~150 articles were affected. I think it's OK to do nothing since the author can archive them manually. What do you think?

PR #1624, opened 2020-11-30
fix(draft): fix self-closing tag exists in draft content

devformatters:
After investigation, found out cheerio will convert iframe into self-closing tag during processing. More info, search self-closing tag in cheerio's Github repository.


guoliu (approved):
Nice fix! 🍃

robertu7 (approved):
Shall we deploy a hotfix?

PR #1654, opened 2020-12-16
Refactor: merge notices that have same entities

robertu7:
**Changes**
* merge notice types that share the same entities; (breaking changes)
* deprecate legacy notice types;
* break down `enums.ts`;

**TODO**
* add circle related notices;


guoliu (approved):
🍄💪

src/common/utils/notice.ts
```diff
@@ -1,106 +1,41 @@
...
-  article_new_subscriber: true,
-  article_new_comment: true,
-  article_mentioned_you: true,
-  subscribed_article_new_comment: true,
-  upstream_article_archived: true,
-  downstream_article_archived: true,
-  comment_pinned: true,
-  comment_new_reply: true,
-  comment_mentioned_you: true,
-  official_announcement: false,
```
devformatters, line 48: This is `false`.

devformatters (approved):
Go. 🧥

src/common/utils/notice.ts
robertu7: nice catch!

PR #1661, opened 2020-12-19
Fix/visitor read record

Carolusian:


devformatters (approved):
Let's try it. 🪡

src/connectors/articleService.ts
```diff
@@ -1290,7 +1290,10 @@ export class ArticleService extends BaseService {
     }
 
     // past record
-    const record = await this.baseFind({ where: { articleId, userId }, table })
+    // if visitor, distinguish base on ip address
+    const record = userId ?
+      await this.baseFind({ where: { articleId, userId }, table }) :
+      await this.baseFind({ where: { articleId, ip }, table })
```
guoliu, line 1296: @Carolusian we are bundling all read records of visitors into a same record, and `ip` here is only recording the latest read. So it seems we shouldn't be finding records based on `ip`, but could do something like this:
```
  const record = userId ?
    await this.baseFind({ where: { articleId, userId }, table }) :
    await this.baseFind({ where: { articleId, userId: null }, table })
```
Which should be the same with the original code then.

guoliu (changes requested):

src/connectors/articleService.ts
Carolusian: @guoliu 

```
  const record = userId ?
    await this.baseFind({ where: { articleId, userId }, table }) :
    await this.baseFind({ where: { articleId, userId: null }, table })
```
is equivalent to 
```
  const record = await this.baseFind({ where: { articleId, userId }, table })
```

The actual code cause the bug is:
```
if (readLength > MINUTE * 30) {
       return { newRead: false }
}
```

If we don't record visitor by IP, I updated the code to:
```
if (userId && readLength > MINUTE * 30) {
       return { newRead: false }
}
```

guoliu: Ah I see, thank you!

guoliu:
Looks good to me 🌳

guoliu (approved):

PR #1664, opened 2020-12-22
Basic db schema for circle

guoliu:
As title. Relationship design can be found in [documentation](https://paper.dropbox.com/doc/6C3tTA88yBMrfM4otFZEf).


db/migrations/20201220213114_create_circle_basic_schema.js
```diff
@@ -0,0 +1,217 @@
...
+      .onUpdate('CASCADE')
+      .onDelete('CASCADE')
+    t.foreign('target_id')
+      .references('id')
+      .inTable('circle')
+      .onUpdate('CASCADE')
+      .onDelete('CASCADE')
+  })
+
+  // article_circle table
```
robertu7, line 66: typo?

db/migrations/20201220213114_create_circle_basic_schema.js
```diff
@@ -0,0 +1,217 @@
...
+  table = 'circle_price'
+
+  await knex('entity_type').insert({ table })
+  await knex.schema.createTable(table, (t) => {
+    t.bigIncrements('id')
+
+    t.string('name').notNullable().unique()
```
robertu7, line 73: `.notNullable` and `.unique` may be unnecessary or we should name it `display_name` instead ?

db/migrations/20201220213114_create_circle_basic_schema.js
```diff
@@ -0,0 +1,217 @@
...
+    t.integer('amount').notNullable()
+    t.enu('state', ['active', 'banned', 'archived'])
```
robertu7, line 75: missing `notNullable()` and `defaultTo`?

db/migrations/20201220213114_create_circle_basic_schema.js
```diff
@@ -0,0 +1,217 @@
...
+
+  // article_circle table
+  table = 'circle_price'
+
+  await knex('entity_type').insert({ table })
+  await knex.schema.createTable(table, (t) => {
+    t.bigIncrements('id')
+
+    t.string('name').notNullable().unique()
+    t.integer('amount').notNullable()
```
robertu7, line 74: should be `decimal` if it's aligned with [transaction](https://github.com/thematters/matters-server/blob/develop/db/migrations/20200407214326_create_transaction_table.js#L14).

db/migrations/20201220213114_create_circle_basic_schema.js
```diff
@@ -0,0 +1,217 @@
...
+    t.enu('state', ['active', 'banned', 'archived'])
+    t.enu('currency', ['HKD', 'LIKE']).notNullable().defaultTo('HKD')
+    t.bigInteger('circle_id').unsigned().notNullable()
```
robertu7, line 77: I'm thinking that it should use `target_id` & `target_type_id` instead and cut the `circle_` prefix from table name if it isn't just for circle. Same for other `circle_` tables.

cc @devformatters

db/migrations/20201220213114_create_circle_basic_schema.js
```diff
@@ -0,0 +1,217 @@
...
+
+  // table for article - circle relationship
+  table = 'article_circle'
+
+  await knex('entity_type').insert({ table })
+  await knex.schema.createTable(table, (t) => {
+    t.bigIncrements('id')
+    t.bigInteger('article_id').unsigned().notNullable()
+    t.bigInteger('circle_id').unsigned().notNullable()
+    t.bigInteger('creator').unsigned().notNullable()
```
robertu7, line 100: what does `creator` stands for?

db/migrations/20201220213114_create_circle_basic_schema.js
```diff
@@ -0,0 +1,217 @@
...
+      .onDelete('SET NULL')
+  })
+
+  // table for user - subscription relationship
+  table = 'circle_subscription'
+
+  await knex('entity_type').insert({ table })
+  await knex.schema.createTable(table, (t) => {
+    t.bigIncrements('id')
+    t.enu('state', ['active', 'archived'])
```
robertu7, line 128: syncing with [Stripe](https://stripe.com/docs/api/subscriptions/object#subscription_object-status)'s?

db/migrations/20201220213114_create_circle_basic_schema.js
```diff
@@ -0,0 +1,217 @@
...
+      .inTable('circle_price')
+      .onUpdate('CASCADE')
+      .onDelete('CASCADE')
+  })
+
+  // update comment table to target circle or article
+  // add columns
+  table = 'comment'
+  await knex.schema.table(table, (t) => {
+    t.bigInteger('target_id').unsigned()
```
robertu7, line 179: `target_id ` and `target_type_id` should be `.notNullable()`?

robertu7 (dismissed):
Indexes are missing, but we can add them back when the API is implemented and the DB usage is clearer.

db/migrations/20201220213114_create_circle_basic_schema.js
```diff
@@ -0,0 +1,217 @@
...
+    t.string('provider_product_id').notNullable().unique()
+    t.timestamp('created_at').defaultTo(knex.fn.now())
+    t.timestamp('updated_at').defaultTo(knex.fn.now())
+
+    // foreign keys
+    t.foreign('owner')
+      .references('id')
+      .inTable('user')
+      .onUpdate('CASCADE')
+      .onDelete('CASCADE')
```
devformatters, line 28: `RESTRICT` ?

devformatters (approved):
Left some comments. 🔭

db/migrations/20201220213114_create_circle_basic_schema.js
guoliu: If restrict, we won't be able to delete an user who has created any circles, for example in implementing user account deletion. However we can replace user data with place holders instead of deleting them. I'll change it to `RESTRICT` for now.

db/migrations/20201220213114_create_circle_basic_schema.js
guoliu: hmmm not sure where the typo is?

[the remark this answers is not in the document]
guoliu: Same as above, not sure where the typo is 😂

db/migrations/20201220213114_create_circle_basic_schema.js
guoliu: This is the name string we need to use in url, so it does need to be notNullable and unique. There is a separate display name field which doesn't need to be unique.

db/migrations/20201220213114_create_circle_basic_schema.js
guoliu: Since for now we can only think of the use case of circle, I think we can reference circle table specifically, and do a migration similar to that of comment table when we need so. Especially when we are not sure how to guarantee referential integrity with `target_id` & `target_type_id`.

db/migrations/20201220213114_create_circle_basic_schema.js
guoliu: Similar with `article_tag`, we can record who add the article to this circle, since we might enable collaboration in the future. But it's currently not needed, removing it for now.

db/migrations/20201220213114_create_circle_basic_schema.js
guoliu: Constraints are added [later](https://github.com/thematters/matters-server/pull/1664/files#diff-86288cb77645b5bc4f723d086767d49b6c583ae98433a8fa83a58c2e3ac7b59bR200) since we need to do migration to fill in data first.

devformatters:
@guoliu Maybe add indexes for tables?

guoliu:
> @guoliu Maybe add indexes for tables?

@devformatters Added, plz review 🙏

devformatters:
LGTM.

db/migrations/20201220213114_create_circle_basic_schema.js
devformatters: And how do we know comments belong to broadcast or discussion?

guoliu: Good point, I think the simplest way is adding a `type` field 🤔

jeremyOk (approved):

PR #1678, opened 2021-01-03
Fix/visitor read count

Carolusian:
@guoliu if we put visitor read count in a separate table, but keep using the same API, it will kinda mix the logic of manipulating two data tables in the same function. So I keep using the same table in this modification and remove `read_time` data for visitors. See if it is OK.


guoliu (changes requested):
@Carolusian Reusing the same table sounds good to me.

I think we also need to use null for read time [here](https://github.com/thematters/matters-server/blob/develop/src/connectors/articleService.ts#L1291) when creating record for visitors. And we should put the update logic for visitors before we [start calculating heartbeat time](https://github.com/thematters/matters-server/blob/develop/src/connectors/articleService.ts#L1300). For both create and update, we should also return `{newRead: true}` since visitor reads are all regarded as separate.

We will also need to update the logic on frontend, so that we skip [the heartbeat timer](https://github.com/thematters/matters-server/blob/develop/src/connectors/articleService.ts#L1300) entirely and only call articleRead once when the viewer is a visitor

guoliu (approved):
Looks good! We probably want to deploy it after the frontend is also ready 🤓 🌴

Carolusian:
> Looks good! We probably want to deploy it after the frontend is also ready 🤓 🌴

@guoliu Guo, the frontend part is done via this PR, and is merged to develop env as well. https://github.com/thematters/matters-web/pull/1720

PR #1679, opened 2021-01-03
feat: add customer portal to user resolver

Carolusian:


src/connectors/stripe/index.ts
```diff
@@ -246,6 +246,24 @@ class StripeService {
...
+   * Get customer portal URL
+   */
+  getCustomerPortal = async ({ customerId }: { customerId: string }) => {
+    try {
+      if (!environment.stripeReturnURL) {
+        throw new ServerError('matters stripe return URL has not been set')
+      }
+      const session = await this.stripe.billingPortal.sessions.create({
+        customer: customerId,
+        return_url: environment.stripeReturnURL,
```
robertu7, line 260: I saw `return_url` is optional and can be configured in [Stripe Dashboard](https://dashboard.stripe.com/settings/billing/portal), would it more simple if use the latter instead?

robertu7 (changes requested):
left a comment about the return URL, others LGTM.

src/connectors/stripe/index.ts
Carolusian: @robertu7 the `return_url` configuration is removed from the code as requested. pls check.

robertu7 (approved):
Let's try. ☃️

PR #1687, opened 2021-01-05
fix(circle): fix circle queries

robertu7:


devformatters:
oh, I might find and replace `circle_price` with `price` in last editting

devformatters (approved):
Go. 👾

PR #1688, opened 2021-01-05
remove `Price.billingCycle`

robertu7:


src/definitions/schema.d.ts
```diff
@@ -1449,7 +1449,7 @@ export const enum GQLTransactionCurrency {
 }
 
 export const enum GQLPriceBillingCycle {
-  month = 'month',
+  monthly = 'monthly',
```
devformatters, line 1452: FYI, Stripe use noun such as `month` and `day`, so I use `month` here instead of `monthly`.

devformatters (approved):
Go. 🦁

guoliu (approved):

PR #1697, opened 2021-01-07
feat(queue): create queue for handling assets

devformatters:


robertu7 (approved):
Since we can figure out which assets are unused through DB, could it be a common cronjob that cleaning them recurringly and automatically?

devformatters:
> Since we can figure out which assets are unused through DB, could it be a common cronjob that cleaning them recurringly and automatically?

It sounds fine to me. We can have one queue job and one cron job to rescan unused assets.

PR #1701, opened 2021-01-08
fix(circle): correct circle subscription input

devformatters:


src/types/circle.ts
```diff
@@ -172,6 +172,11 @@ export default /* GraphQL */ `
     targetId: ID!
   }
 
+  input CircleSubscriptionInput {
+    "Unique ID."
+    id: ID!
```
robertu7, line 177: missing `password: String` for further subscriptions?

robertu7: related discussion: https://mattersnews.slack.com/archives/C01CYECRYTH/p1609814350006600

robertu7 (changes requested):

src/types/circle.ts
devformatters: Added. Thanks.

src/types/circle.ts
```diff
@@ -172,6 +172,19 @@ export default /* GraphQL */ `
     targetId: ID!
   }
 
+  input SubscribeCircleInput {
+    "Unique ID."
+    id: ID!
+
+    "Wallet password."
+    password: String!
```
robertu7, line 180: Is password required for the first subscription?

robertu7 (approved):

src/types/circle.ts
devformatters: I'll merge it first, and then we can discuss it in the following works.

PR #1713, opened 2021-01-12
feat(payment pointer): database, mutation and query for payment pointer

guoliu:
* add database, mutation and query for payment pointer
* pull out some reusable functions from userService to local directory


src/mutations/user/utils/updateDbEs.ts
```diff
@@ -0,0 +1,56 @@
...
+import logger from 'common/logger'
+import { es, knex } from 'connectors'
+import { GQLUpdateUserInfoInput, UserRole } from 'definitions'
+
+/**
+ * Update user db & es record
+ * @param id: user id
+ * @param data: data to be updated
+ */
+export const updateDbEs = async (
```
devformatters, line 12: To me, `utils` is more like a place for functions of conversion, formatting and calculation. I recommend to drop function that causes side effect into atomService. I might put es update function in atomService, and reuse `atomService.update` and `atomService.updateEs` in resolvers.

devformatters (approved):
Go. 👨🏾‍✈️

src/mutations/user/utils/updateDbEs.ts
guoliu: @devformatters updated as you suggest, plz review again 🙏

jeremyOk (approved):

PR #1714, opened 2021-01-13
Circle Draft

robertu7:
* Add support for adding a circle to draft and article;
* Add support for customizing summary of draft & article;
* Add `CIRCLE_NOT_FOUND` error;

### DIFF
--- a/db/migrations/20210113172906_add_circles_to_draft.js
+++ b/db/migrations/20210113172906_add_circles_to_draft.js
@@ -1,0 +1,13 @@
+const table = 'draft'
+
+exports.up = async (knex) => {
+  await knex.schema.table(table, function (t) {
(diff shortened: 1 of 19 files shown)


db/migrations/20210113172906_add_circles_to_draft.js
```diff
@@ -0,0 +1,13 @@
+const table = 'draft'
+
+exports.up = async (knex) => {
+  await knex.schema.table(table, function (t) {
+    t.specificType('circles', 'text ARRAY')
```
guoliu, line 5: I think the product logic we have now is one draft can only have one circle. This is quite unlikely to change in the future. Do we still need to use array here for future proof?

guoliu (approved):
Left a comment but looks good to me 🌾

PR #1727, opened 2021-01-19
feat(draft): add `summaryCustomized` to draft & article;

robertu7:
Also add test cases for `putDraft` and `editArticle`;


guoliu (changes requested):
Shall we just use `summary` in db schema and API, and overwrite this field when user input their summary? If we can treat the default summary and customized summary the same way, we can simplify the overall logic.

robertu7:
> Shall we just use `summary` in db schema and API, and overwrite this field when user input their summary? If we can treat the default summary and customized summary the same way, we can simplify the overall logic.

We will be showing the summary in the article detail only if it's customized, relying on `summary` is hard to do that.

image

guoliu (approved):

guoliu:
> > Shall we just use `summary` in db schema and API, and overwrite this field when user input their summary? If we can treat the default summary and customized summary the same way, we can simplify the overall logic.
> 
> We will be showing the summary in the article detail only if it's customized, relying on `summary` is hard to do that.
> 
> image

Sounds good, thanks for the explanation!

devformatters (approved):
Go. ✊🏻

PR #1734, opened 2021-01-22
feat(stripe): add slack notifier for Stripe

devformatters:


src/connectors/queue/stripe.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  private syncDeliveryFailedEvents: Queue.ProcessCallbackFunction<
+    unknown
+  > = async (job, done) => {
+    try {
+      logger.info('[schedule job] sync delivery failed events')
+
+      // query delivery failed events
+      const result = await this.paymentService.stripe.getDeliveryFailedEvents()
```
robertu7, line 53: Only retrieving events for the past 24 hours looks more polite if failed events can't be resolved on time.

src/routes/pay/stripe.ts
```diff
@@ -179,39 +180,57 @@ stripeRouter.post('/', async (req, res) => {
...
+        slack.sendStripeAlert({
+          data: event,
+          message: 'Unexpected event type',
+        })
+        break
+    }
+  } catch (error) {
+    logger.error(error)
+    slack.sendStripeAlert({
+      data: event,
```
robertu7, line 230: Should  we pass event data like [queue does](https://github.com/thematters/matters-server/pull/1734/files#diff-4976b7a1365825fb091212aef498537e539e36c5da1b9ed24b21f6a29d805788R60-R64) since the event may contain sensitive data?

robertu7 (approved):

guoliu (approved):
🌴

src/routes/pay/stripe.ts
devformatters: Revised. Thanks.

src/connectors/queue/stripe.ts
devformatters: Adjusted the interval of repeat job from daily to 3hrs

PR #1741, opened 2021-01-26
Circle Subscription

robertu7:
* Subscribe Circle via password or credit card:
  * For password, the subscription will be settled synchronously;
  * For credit card, the subscription will be settled via Webhook asynchronously;
* Unsubscribe Circle
  * Cancel subscription if it only has one subscription item; (removed Matters price placeholder);
  * Archive DB subscription item instead of deleting it for keeping the subscription records;
* Webhook to sync subscription and subscription items;


jeremyOk (approved):

devformatters (approved):
Go. 👂🏻

PR #1769, opened 2021-02-18
fix(circle): fix circle pinned broadcast order

robertu7:


src/queries/comment/circle/pinnedBroadcast.ts
```diff
@@ -14,7 +14,7 @@ const resolver: CircleToPinnedBroadcastResolver = async (
       targetId: id,
       type: COMMENT_TYPE.circleBroadcast,
     },
-    orderBy: [{ column: 'created_at', order: 'desc' }],
+    orderBy: [{ column: 'updated_at', order: 'desc' }],
```
guoliu, line 17: If we use `updated_at` to order, the whenever a comment is edited it will be ranked first. Is that intended? 🤔

devformatters (approved):
Most looks good except the order column Guo pointed out

guoliu (approved):

jeremyOk (approved):

PR #1782, opened 2021-02-25
Revise Circle Notices

robertu7:
* remove `CircleNewDiscussion` notice; 
* add `circle_*_mentioned_you`, `circle_*_new_reply` DB notice types to distinguish from article comment since some components (daily summary email) rely on it;


guoliu (approved):
🍀

PR #1790, opened 2021-03-02
Fix incorrect order of dataloader

robertu7:
Same as https://github.com/thematters/matters-server/blob/develop/src/connectors/baseService.ts#L64


devformatters (approved):
Go. 👾

PR #1793, opened 2021-03-04
fix(circle): return empty discussion for non-member

robertu7:
* Fix article content returns an empty string for circle member;
* Revise name validators;

### DIFF
--- a/src/queries/comment/circle/discussion.ts
+++ b/src/queries/comment/circle/discussion.ts
@@ -1,14 +1,40 @@
-import { COMMENT_STATE, COMMENT_TYPE } from 'common/enums'
+import {
+  COMMENT_STATE,
+  COMMENT_TYPE,
+  PRICE_STATE,
+  SUBSCRIPTION_STATE,
+} from 'common/enums'
 import { connectionFromArray, cursorToIndex } from 'common/utils'
 import { CircleToDiscussionResolver } from 'definitions'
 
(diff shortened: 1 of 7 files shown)


src/queries/comment/circle/discussion.ts
```diff
@@ -1,16 +1,42 @@
...
   { id },
   { input },
-  { dataSources: { atomService } }
+  { viewer, dataSources: { atomService }, knex }
 ) => {
   if (!id) {
     return connectionFromArray([], input)
   }
 
+  const records = await knex
```
guoliu, line 19: Would it be more consistent to use knex in `atomService`?

guoliu (approved):

src/queries/comment/circle/discussion.ts
robertu7: It seems hard to use `atomService` since we use `join` here, but there are many places that have circle member query, maybe we should consider how to reuse it.

devformatters: It's quite different and difficult to simulate Prisma-like `join` feature using Knex-like functions. So, complex operations currently are not implemented in atomService. To me, it's fine to use `knex` here, and I believed it will replace by Prisma's way in the future.

robertu7: Just moved to `paymentService` for reusability.

PR #1794, opened 2021-03-05
feat(circle): add `discussionCount` and `discussionThreadCount`

robertu7:
Since circle discussion are protected and `discussion.totalCount` returns `0` on public query, we add `discussionCount` and `discussionThreadCount` to fulfill the stats needs.


devformatters (approved):
Go. 👨🏼‍🔬

PR #1843, opened 2021-03-15
Change the fetched size of events of stripe alert

devformatters:


src/connectors/stripe/index.ts
```diff
@@ -361,11 +363,42 @@ class StripeService {
...
+
+      const now = Date.now()
+      const week = DAY * 7
+      const events: Array<Record<string, any>> = []
+
+      while (hasMore && fetch) {
+        // fetch events from stripe
+        const batch: Record<string, any> = await this.stripeAPI.events.list({
+          delivery_success: false,
+          limit: 10,
```
robertu7, line 383: To reduce the API requests, should we increase the `limit` to the upper (100)?

robertu7 (approved):
🚼

src/connectors/stripe/index.ts
devformatters: 👌🏻 Resize to 50.

PR #1857, opened 2021-03-23
feat: apply coupon to circle subscription

Carolusian:


devformatters (approved):
Let's try it. 🖇

guoliu (approved):

src/connectors/paymentService.ts
```diff
@@ -695,11 +696,21 @@ export class PaymentService extends BaseService {
...
     const { userId, priceId, providerCustomerId, providerPriceId } = data
 
+    const ivt = await this.findPendingInvitation({ userId, priceId })
     // Create from Stripe
-    const stripeSubscription = await this.stripe.createSubscription({
-      customer: providerCustomerId,
-      price: providerPriceId,
-    })
+    let stripeSubscription
+    if (isUndefined(ivt)) {
```
robertu7, line 702: it's ok to just `ivt` instead of `isUndefined(ivt)`

robertu7 (approved):

src/connectors/paymentService.ts
```diff
@@ -760,5 +786,50 @@ export class PaymentService extends BaseService {
...
+  findPendingInvitation = async (params: {
+    userId: string
+    priceId: string
+  }) => {
+    const records = await this.knex
+      .select('ci.id', 'cc.provider_coupon_id')
+      .from('circle_invitation as ci')
+      .join('circle_price as cp', 'cp.circle_id', 'ci.circle_id')
+      .join('circle_coupon as cc', 'cc.id', 'ci.coupon_id')
+      .where({
```
robertu7, line 813: Since we allow circle owners to invite via email, we should find invitation via `user.email` also?

robertu7 (changes requested):

src/connectors/paymentService.ts
Carolusian: merged for the time being, I will have another PR for this.

Carolusian: @robertu7 please see the updated where clause for email checking

Carolusian (commented):
@robertu7 I removed `isUndefined` from paymentService. But opted to keep it in stripe/index.ts as it seems more explicit in optional parameter checking.

robertu7 (approved):

src/connectors/paymentService.ts
devformatters: Nice point. 💪🏻

PR #1892, opened 2021-04-01
fix(time): fix time calculation of `getUTC8NextMonthDayOne`; add test cases for time utils;

robertu7:


src/common/utils/time.ts
```diff
@@ -39,15 +37,17 @@ export const getUTC8NextMonthDayOne = () => {
     date.setMonth(month + 1)
   }
 
-  // set date
-  date.setDate(1)
+  // set year
+  if (isLastMonthOfYear) {
+    date.setFullYear(date.getFullYear() + 1)
+  }
```
devformatters, line 43: Nice!

devformatters (approved):
Go. ✊🏻

PR #1900, opened 2021-04-07
fix(editor): correct nested br tags in article content

devformatters:


robertu7 (approved):
Shall we add test cases for `correctSelfClosingHtmlTag` and `correctNestedBrTag`?

devformatters:
> Shall we add test cases for `correctSelfClosingHtmlTag` and `correctNestedBrTag`?

Must have. I've added. 🙌

PR #1904, opened 2021-04-08
feat(collection): disallow blocked user to collecting articles

devformatters:


src/mutations/article/editArticle.ts
```diff
@@ -219,9 +220,47 @@ const resolver: MutationToEditArticleResolver = async (
...
+            })
+
+            if (isBlocked) {
+              return
+            }
+
+            return articleDbId
+          })
+        )
+      ).filter((articleId): articleId is string => !!articleId)
```
robertu7, line 261: `articleId is string` new feature on TypeScript? 👀

robertu7 (approved):

src/mutations/article/editArticle.ts
devformatters: It's not a new feature. The concept is similar to `Type guard`.

PR #1942, opened 2021-04-20
New Cross Border Payout Implementation

robertu7:
* Onboard Stripe Connect account with service agreement based on country;
* Use `transfer` API instead of destination charge;


src/connectors/stripe/index.ts
```diff
@@ -40,7 +42,7 @@ class StripeService {
     }
 
     this.stripeAPI = new Stripe(environment.stripeSecret, {
-      apiVersion: '2020-03-02',
+      apiVersion: '2020-08-27',
```
devformatters, line 45: 👍🏻

devformatters (approved):
Go. 🦾

PR #1945, opened 2021-04-21
feat(cover): Add the logic of fallback cover when drafting

devformatters:


robertu7 (approved):
a little bit lost by the nested if statements, let's try anyway. XD

guoliu (approved):

PR #1961, opened 2021-04-26
Matters as subscription provider

robertu7:
In this PR, we add support for creating subscription and subscription items with `provider=matters`, which currently used by trial invitation.

**Notes:**
* Matters subscription items also count in the limitation that up to 20;
* `freePeriod` should be altered from month to day unit by the next deployment (June 1);


guoliu (approved):

devformatters (approved):
Go. 🎃

PR #1965, opened 2021-04-27
feat(migrate): add migration script for Matters coupon

devformatters:


db/migrations/20210427123109_migrate_subscription.js
```diff
@@ -0,0 +1,193 @@
...
+  const update = async ({ table, where, data }) => {
+    const [record] = await knex(table).where(where).update(data).returning('*')
+    return record
+  }
+
+  // fetch accepted invitations
+  const invts = await knex
+    .select()
+    .from(t_invitation)
+    .where({ accepted: true })
```
robertu7, line 45: skip migrated invitations (`subscription_item_id` ) if we retry?

db/migrations/20210427123109_migrate_subscription.js
```diff
@@ -0,0 +1,193 @@
...
+
+    const user = await findFirst({
+      table: t_user,
+      where: { ...(userId ? { id: userId } : { email }) },
+    })
+
+    if (!user) {
+      continue
+    }
+
```
robertu7, line 63: Is knex migration transactional (one failure of invitation migration will fail them all)? Shall we print debug info with masked user data (e.g.user id) on every step?

db/migrations/20210427123109_migrate_subscription.js
```diff
@@ -0,0 +1,193 @@
...
+          provider: 'matters',
+          provider_subscription_id: v4(),
+          state: 'trialing',
+          user_id: user.id,
+        },
+      })
+    }
+
+    // Step 3: create Matters subscription item
+    const { id, circle_id, ...rest } = item
```
robertu7, line 112: explicit better than implicit (`...rest`)?

robertu7 (changes requested):

guoliu (approved):
Let's try it 👍

db/migrations/20210427123109_migrate_subscription.js
devformatters: Added.

db/migrations/20210427123109_migrate_subscription.js
devformatters: Added logger and break original script into two migration.

db/migrations/20210427123109_migrate_subscription.js
devformatters: Changed.

db/migrations/20210427142108_migrate_stripe_subscription.js
```diff
@@ -0,0 +1,100 @@
...
+
+      if (!stripeSubscription) {
+        console.log('Stripe subscription not found')
+        continue
+      }
+
+      if (stripeSubscription.canceled_at > 0) {
+        console.log('Stripe subscription arleady canceled')
+        continue
+      }
```
robertu7, line 68: also need to check if subscription items already been deleted?

robertu7 (approved):

db/migrations/20210427142108_migrate_stripe_subscription.js
devformatters: Nice point. Added.

robertu7 (approved):

PR #1989, opened 2021-05-04
feat(env): local dev compatible with GitHub Workspace

robertu7:


src/index.ts
```diff
@@ -31,6 +31,7 @@ try {
 const PORT = 4000
 const app = express()
 const httpServer = http.createServer(app)
+app.set('trust proxy', ['loopback', 'linklocal', 'uniquelocal'])
```
devformatters, line 34: 🤔

devformatters (approved):
Go.

src/index.ts
robertu7: In GitHub Workspace with port forwarding, `req.hostname` of API requests is `localhost` when calling `https://robertu7-thematters-matters-server-hj5h-4000.githubpreview.dev/graphql`, and which isn't what we expected. 

related doc: http://expressjs.com/en/guide/behind-proxies.html

robertu7: These changes need a test on local and develop environments.

PR #2003, opened 2021-05-07
feat(circle): only open customer portal if viewer has non-canceled subscriptions

robertu7:


devformatters (approved):
Go. 🎒

src/queries/user/wallet/customerPortal.ts
```diff
@@ -1,11 +1,55 @@
...
+      archived: false,
+    })
+    .first()) as Customer
+
+  if (!customer) {
+    return null
+  }
+
+  const customerId = customer.customerId
+  return paymentService.stripe.getCustomerPortal({ customerId })
```
devformatters, line 52: Nice. it's built-in.

PR #2024, opened 2021-05-14
fix(content): revise util of correcting nested BR tags in html string

devformatters:


robertu7 (approved):
👷have to say it's a tough issue.

devformatters:
> 👷have to say it's a tough issue.

🦊  🦊  🦊  🦊

PR #2041, opened 2021-05-20
feat(invites): add new query fields for invites

devformatters:


schema.graphql
```diff
@@ -986,8 +989,14 @@ type Invitation {
   """Sent time."""
   sentAt: DateTime!
 
+  """Accepted time."""
+  acceptedAt: DateTime
+
   """Determine it is accepted or not."""
   accepted: Boolean!
```
robertu7, line 996: looks `accepted` can be deprecated

robertu7 (approved):
🍉

schema.graphql
devformatters: Revised. Thanks.

PR #2055, opened 2021-05-31
fix(cookie): set samesite to `none` if it's local request

robertu7:
Chrome 91 may change the default behavior of cookie `SameSite`, we need to set it to `none` explicitly.


devformatters (approved):
Go. 🥽

PR #2064, opened 2021-06-03
fix(access): revise schema of article access

devformatters:


src/queries/article/access/secret.ts
```diff
@@ -2,17 +2,27 @@ import { CIRCLE_STATE } from 'common/enums'
...
-  { knex }
+  { viewer, knex }
 ) => {
+  if (!authorId || !viewer.id) {
+    return
+  }
+
+  // check viewer is owner
+  if (authorId !== viewer.id) {
+    return
```
guoliu, line 15: Shall we throw an error instead of returning null? Returning null might be cached, then when the author query the secret s/he gets null as well

robertu7: should be fine since caching is based on `viewer.id`?

robertu7 (approved):

src/queries/article/access/secret.ts
devformatters: Fixed it.

PR #2078, opened 2021-06-11
Update articleService.ts

jeremyOk:
It seems that the condition showed[ in this line](https://github.com/thematters/matters-server/blob/a5e1a7aa8419146158d6e416a859ae5bff613ace/src/connectors/articleService.ts#L1396) is wrong. It should be `if (userId && readLength < MINUTE * 30)`


src/connectors/articleService.ts
```diff
@@ -1385,27 +1385,33 @@ export class ArticleService extends BaseService {
...
 
     // calculate last read total time
     const readLength = Date.now() - new Date(oldData.lastRead).getTime()
 
-    // if original read longer than 30 minutes
-    // skip
-    if (userId && readLength > MINUTE * 30) {
+    // if lapse shorter than 5 minutes or readLength longer than 30 minutes, 
+    // only update the updatedAt and then return
+   if (lapse < MINUTE * 5 || readLength > MINUTE * 30) {
```
guoliu, line 1396: Should be `lapse < MINUTE * 5 && readLength > MINUTE * 30`, since we want to stop recording when the accumulated time is larger than 30 minutes. If we stop recording when lapse is smaller than 5 minute **or** when  accumulated time is larger than 30 minutes, then it will never record since normally there's less than 5 minutes lapse between heartbeats (frontend calls every 5 seconds).

src/connectors/articleService.ts
```diff
@@ -1385,27 +1385,33 @@ export class ArticleService extends BaseService {
...
       return { newRead: false }
-    }
+    } 
 
-    // if lapse is longer than 5 minutes,
-    // or total length longer than 30 minutes,
+    // If lapse equal or longer than 5 minutes,
     // add a new count and update last read timestamp
-    if (lapse > MINUTE * 5 || readLength > MINUTE * 30) {
+   if (lapse >= MINUTE * 5 && readLength <= MINUTE * 30) {
```
guoliu, line 1409: I think here is the case where we treat the reading as a new read, which corresponds to the condition where lapse is larger than 5 minutes; meaning the time between last heartbeat and this one is longer than 5 minutes, so it is a different read.

src/connectors/articleService.ts
```diff
@@ -1385,27 +1385,33 @@ export class ArticleService extends BaseService {
...
-    // if original read longer than 30 minutes
-    // skip
-    if (userId && readLength > MINUTE * 30) {
+    // if lapse shorter than 5 minutes or readLength longer than 30 minutes, 
+    // only update the updatedAt and then return
+   if (lapse < MINUTE * 5 || readLength > MINUTE * 30) {
+      await this.baseUpdate(
+          oldData.id,
+          {
+            readTime
```
guoliu, line 1400: `readTime` is defined on [line 1416](https://github.com/thematters/matters-server/pull/2078/files#diff-6775d7dc5f0198959c6fe0e20d1aeff1620ba32b75bb328a09c7d042230da5c9R1416), so we need to move the definition before this line.

guoliu:
Merging to branch https://github.com/thematters/matters-server/tree/fix/read-time to add more changes.

PR #2111, opened 2021-07-01
WIP: fix(code): better error code returning in confirmation

devformatters:


devformatters:
We need a better flow to code verification to tackle this.

PR #2122, opened 2021-07-07
Implementation of New Following Feed

robertu7:
**Changes**
* Add two materialized views for global user activities in different needs:
  * `user_activity_materialized`, 30 days window, refresh every 3 minutes;
  * `user_activity_long_materialized`, 90 days window, refresh every hour;
* Add `createdAt` to `FollowingActivity`;

**Notes**
* Time window is to be determined;
* `user_activity_long_materialized`, as a fallback of `user_activity_materialized`, isn't in use currently, also need some inputs based on current implementation;


schema.graphql
```diff
@@ -1692,7 +1692,7 @@ type RecentSearchEdge {
 
 type Recommendation {
   """Activities based on user's following, sort by creation time."""
-  following(input: ConnectionArgs!): FollowingActivityConnection! @deprecated(reason: "Merged into `Recommendation.following`")
```
robertu7, line 1695: typo from previous PR

devformatters (approved):
Really nice work. Let's try it out.

guoliu (approved):

PR #2173, opened 2021-07-22
fix(ci): fix wrong ci syntax of schema push step

robertu7:
Also,
* remove `@constraint ` on `type` field since we use it only on `input` for now.


devformatters (approved):
Go. 👩🏻‍🚀

PR #2209, opened 2021-08-09
feat(following): alter "recommended_articles_from_read_tags_materialized"

robertu7:
* Exclude user's own articles;
* Exclude articles state is not `active`;
* Exclude articles author state is not `active;
* Remove `article_prepared_limit`;
* Shorten refresh interval;


guoliu (approved):
Let's try it

devformatters (approved):
Go. 🛰

PR #2227, opened 2021-08-16
[hotfix -master] fix(playground): fix playground inaccessibility in production by disabling csp

robertu7:


devformatters (approved):
Go. 🚩

PR #2228, opened 2021-08-17
Announcement management APIs

devformatters:


jeremyOk (approved):

src/types/system.ts
```diff
@@ -224,6 +242,25 @@ export default /* GraphQL */ `
...
+  input AnnouncementsInput {
+    id: ID
+  }
+
+  input PutAnnouncementInput {
+    id: ID
+    title: String
+    cover: String
+    content: String
+    link: String
```
robertu7, line 254: use `@constraint`?

src/types/system.ts
```diff
@@ -43,17 +45,33 @@ export default /* GraphQL */ `
...
 
+  type Announcement {
+    id: ID!
+    title: String
+    cover: String
+    content: String
+    link: String
+    type: AnnouncementType!
+    visible: Boolean!
+    order: Int!
```
robertu7, line 70: where's the `order` uses for? looks `official/announcements` don't order by `order`.

robertu7 (approved):

src/types/system.ts
devformatters: Fixed. Thanks.

src/types/system.ts
devformatters: It's a optional field for future implementation.

PR #2235, opened 2021-08-19
fix(permission): forbid to comment, subscribe, appreciate or read article which state isn't active

robertu7:


devformatters (approved):
Go. 🍷

PR #2278, opened 2021-09-29
document assets is accessible to author only

tx0c:
> Only the author can access the Article.assets, it's for article edit mode currently

https://github.com/thematters/matters-web/pull/2260#discussion_r718376301


devformatters (approved):
Go. 🛰

PR #2285, opened 2021-09-30
Fix/xss sanity check

tx0c:
- make unittest for server side for sure not to touch the pre code-block … 0341e8f
  part of reason for thematters/matters-web#2114 is server not saving exactly same pre code-block, as

  ```html
  <pre class="ql-syntax" spellcheck="false">Pre1\n</pre><p>next paragraph...
  ```

  server side is always dropping the `spellcheck="false"`, then next time when editor is loading
the incomplete pre code-block, it triggers `clipboard.convert` to parse again, and in MattersArticleEditor
the `mentionContainer` reference is causing ReactQuill to re-initialize twice every time, caused
converting between `HTML <=> Delta` happening 6 or even more times, and eventually caused wrong parsing

- sanitize xss filter whitelist pre: ['spellcheck'] 3c9fccc


guoliu (approved):
Nice catch! 🌿

PR #2286, opened 2021-10-01
Revert "Fix/xss sanity check"

tx0c:
trying to Revert thematters/matters-server#2285 let CI check last deployment failed on unittest why https://github.com/thematters/matters-server/runs/3763557782?check_suite_focus=true


tx0c:
this revert only proves the latest deployment failure is not caused by PR #2285 ; 

see #2287 for the root cause

PR #2299, opened 2021-10-19
API schema for query user NFT ownership

tx0c:
for #2293

the opensea api has more fields for each NFT token, it depends on how much information do we want to show for each token?

    // https://api.opensea.io/api/v1/assets?owner=0xf4b4a58974524e183c275f3c6ea895bc2368e738&limit=2


devformatters (dismissed):

src/types/user.ts
```diff
@@ -516,6 +516,23 @@ export default /* GraphQL */ `
     id: ID!
     address: String!
     createdAt: DateTime!
+
+    """NFT assets owned by this wallet address"""
+    assets: [NFTAsset!]!
```
robertu7, line 521: * `nfts` instead of `assets`?
* `[NFTAsset!]` instead of `[NFTAsset!]!`?

src/types/user.ts
```diff
@@ -516,6 +516,23 @@ export default /* GraphQL */ `
...
+  }
+
+  """ NFT Asset """
+  type NFTAsset {
+    tokenId: String!
+    name: String!
+    description: String
+    imageUrl: String!
```
robertu7, line 529: what are the differences among `image*`? I think we can just keep a thumb/preview (for client-side preview) and an original (for upload back to us via `singleUploadFile` mutation)?

src/types/user.ts
```diff
@@ -516,6 +516,23 @@ export default /* GraphQL */ `
...
+    imagePreviewUrl: String
+    imageThumbnailUrl: String
+    imageOriginalUrl: String!
+    contractAddress: String!
+    collectionName: String!
+    tokenMetadata: String
```
robertu7, line 535: We should strong type `tokenMetadata` if it's a key-value JSON string?

e.g.
```
tokenMetadata: [NFTMetaDataItem!]

type NFTMetadataItem {
    name: String!
    value: String
}
```

src/types/user.ts
```diff
@@ -516,6 +516,23 @@ export default /* GraphQL */ `
...
     address: String!
     createdAt: DateTime!
+
+    """NFT assets owned by this wallet address"""
+    assets: [NFTAsset!]!
+  }
+
+  """ NFT Asset """
+  type NFTAsset {
+    tokenId: String!
```
robertu7, line 526: need a `id` field? `tokenId` may not be unique.

robertu7 (dismissed):

src/types/user.ts
tx0c: sample from https://rinkeby-api.opensea.io/api/v1/assets?asset_contract_address=0xae89d81ab5c668661200fa9c6ed45fe1707f7097&owner=0x08f3BCa6DeF9CdF5EeE1114Ed4d2A0d91cFE4FFA&order_direction=desc&offset=0&limit=20

image_original_url is ipfs original, image_url is cached original size, preview is 250w, thumbnail is 128w

```json
    "image_url": "https://lh3.googleusercontent.com/2D6pKJyKZQ5-FfvwsgofxJp2zEuhEji7iI06Odz-NLqYK1zaQvEbIe0R1M-refwv6la8acbA6ap7k5PxtkbgFNkZHe_eUsl3Wvh9mw",
    "image_preview_url": "https://lh3.googleusercontent.com/2D6pKJyKZQ5-FfvwsgofxJp2zEuhEji7iI06Odz-NLqYK1zaQvEbIe0R1M-refwv6la8acbA6ap7k5PxtkbgFNkZHe_eUsl3Wvh9mw=s250",
    "image_thumbnail_url": "https://lh3.googleusercontent.com/2D6pKJyKZQ5-FfvwsgofxJp2zEuhEji7iI06Odz-NLqYK1zaQvEbIe0R1M-refwv6la8acbA6ap7k5PxtkbgFNkZHe_eUsl3Wvh9mw=s128",
    "image_original_url": "https://ipfs.io/ipfs/QmQvL122xLjLDWwjAJjRNkKWQsh9GzQnkWU5mtc2FPgKnY",
```

src/types/user.ts
tx0c: it can always return a `[]` instead of null, might be helpful for frontend sometimes

src/types/user.ts
tx0c: it's a link actually,
```json
            "token_metadata": "https://ipfs.io/ipfs/QmY5cedmZ27MULssFNnVPtYXwA8uYXUY8BiSaBgDkZ98cR/124",
```

may want the traits?
```json
            "traits": [
                {
                    "trait_type": "Ideology",
                    "value": "Epicurean",
                    "display_type": null,
                    "max_value": null,
                    "trait_count": 3,
                    "order": null
                },
                {
                    "trait_type": "Race",
                    "value": "Valya",
                    "display_type": null,
                    "max_value": null,
                    "trait_count": 25,
                    "order": null
                },
                {
                    "trait_type": "Character",
                    "value": "introverted",
                    "display_type": null,
                    "max_value": null,
                    "trait_count": 8,
                    "order": null
                }
            ],

```

tx0c:
for @robertu7 @devformatters ready review again,

from here https://rinkeby-api.opensea.io/api/v1/assets?asset_contract_address=0xae89d81ab5c668661200fa9c6ed45fe1707f7097&owner=0x0Ec18c1D8df059422101dada0e96bB6D24e7877E&order_direction=desc&offset=0&limit=20 is there any other fields we want to show?

src/types/user.ts
guoliu: I think our design doesn't include traits now, so should be fine without it.

guoliu (approved):
Let's test it out. Not clear whether we should cache the image result or not, will continue discussion on slack.

PR #2327, opened 2021-11-09
fix(deps): move rover installation to CI

robertu7:
As title, M1 (arm) machine cannot install rover in local, move it CI since it's only used by CI.


jeremyOk (approved):

devformatters (approved):
Nice. In fact, I just faced this issue when reviewing challenge codes. 🪱

PR #2329, opened 2021-11-09
feat(wallet): add `user_id` column to `crypto_wallet_signature`

robertu7:


jeremyOk (approved):

src/mutations/user/putWallet.ts
```diff
@@ -96,7 +96,7 @@ const resolver: MutationToPutWalletResolver = async (
     wallet = await atomService.update({
       table,
       where: { id, userId: viewer.id },
-      data: { address },
+      data: { address, updatedAt: new Date() },
```
devformatters, line 99: `where : { ..., archived: false }` ?

devformatters (approved):
Nice. work. 🐷

src/mutations/user/putWallet.ts
robertu7: already did the [checking before](https://github.com/thematters/matters-server/blob/151ce56079d809497ead73fe2fc6f7ed04b62af1/src/mutations/user/putWallet.ts#L87-L94), should be good?

devformatters: but this is going to update both `archived = true or archived = false`

PR #2347, opened 2021-11-24
remove support for get token user via uuid

robertu7:
https://github.com/thematters/matters-server/pull/2146


guoliu (approved):
🍀 although test failed

PR #2348, opened 2021-11-28
WIP: update rating-limit-connections for amazon linux 2

Carolusian:


robertu7:
Since our domain already migrated to Cloudflare with more control on network traffic, I think we could remove related configs on EC2 side.

PR #2371, opened 2021-12-06
fix: avoid the scenario where current folder does not exist

Carolusian:
Fix the following:

- In new EB env, the `current` folder maybe absent during initialization, e.g. when `clone` an environment;
- In new EB env, `nginx.conf` in `.ebextension` does not work.


guoliu (approved):
Let's try it

PR #2405, opened 2021-12-21
fix(likecoin): add logs to debug webhook callback

Carolusian:


src/routes/pay/likecoin.ts
```diff
@@ -215,14 +215,25 @@ likecoinRouter.post('/', async (req, res, next) => {
...
     if (!trans) {
-      throw new Error(`counld not find tx hash`)
+      trans = (
+        await paymentService.findTransactions({
+          providerTxId: txState,
+        })
+      )[0]
+
+      if (!trans) {
+        throw new Error(`counld not find tx hash`)
```
tx0c, line 235: may define a custom error type in `src/common/errors.ts` and use it here?   will be better to find it in Sentry for troubleshooting?

Carolusian: @tx0c good suggestion. let me review the whole file and refactor the error handling accordingly after this fix.

robertu7 (approved):

PR #2419, opened 2022-01-18
default displayName to ethAddress

tx0c:


src/mutations/user/walletLogin.ts
```diff
@@ -169,6 +169,7 @@ const resolver: MutationToWalletLoginResolver = async (
   // Signup otherwise
   const newUser = await userService.create({
     userName: ethAddress,
+    displayName: ethAddress,
```
robertu7, line 172: how about `0x123...abc` as default userName and displayName?

 cc @devformatters

robertu7 (approved):

PR #2422, opened 2022-01-24
feat(wallet): force wallet signup with email

robertu7:
* force wallet signup with email;
  * revert changes on `changeEmail` mutation;
* add `User.info.isWalletAuth` for client to detect if disable reset password feature;
* add `ethAddress` argument to root `user` query for client to detect if current auth is login or sign up;

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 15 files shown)


devformatters (approved):
Nice. 🏓

PR #2434, opened 2022-01-29
use knex builtin `.clone` instead of lodash/clone

### LINKED ISSUE #2433: users' comment page seems showing only last a few, no infinite scrolling
seems to be caused by knex upgrade:

https://github.com/thematters/matters-server/compare/v3.31.0...master?expand=1#diff-c9977537d87d8c449ed2157f27fdb3cf3c29dfa30e7f3b77efa468fb83c88ed0

tx0c:
resolves #2433


devformatters (approved):
Nice. 🎅

PR #2447, opened 2022-02-15
fix/crypto_wallet: fake a table entry if has eth_address linked in us…

tx0c:
…er table

supplements #2429

the old crypto_wallet table could potentially support multiple wallet for single user,
but new model is strictly 1-to-1 mapping;

the `crypto_wallet` table will eventually be deprecated

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -2528,7 +2528,6 @@ type Following {
 type CryptoWallet {
   id: ID!
   address: String!
-  createdAt: DateTime!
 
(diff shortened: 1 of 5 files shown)


src/types/__test__/user.test.ts
```diff
@@ -945,6 +945,7 @@ describe('crypto wallet', () => {
       query: GET_VIEWER_CRYPTO_WALLET,
       variables: {},
     })
+    console.log('viewer.info:', _get(data, 'viewer.info'))
```
guoliu, line 948: Shall we remove this?

guoliu (approved):
LGTM (except a leftover console log)

PR #2452, opened 2022-02-25
fix/wallet-login: should allow re-login of same account

tx0c:
when viewer.ethAddress is not null, this should not go into wallet link logic;

https://auth.matters.news/login?... has a logic to re-login when linking to like.co,
wallet re-login should work same as email re-login: ignore whatever in the viewer

also archive the crypto_wallet at time of wallet link,
supercedes #2450


robertu7 (approved):
🍶

PR #2465, opened 2022-03-15
fix(revise-time): skip the first draft, only return when more than 2 …

tx0c:
…drafts

the first draft always exists; performance fix #2462


guoliu (approved):
THX!

PR #2473, opened 2022-03-17
fix(): change QUEUE_JOB.activateOnboardingUsers frequency

tx0c:


robertu7 (approved):
I have changed to 20 mins, you want to make it longer? 2 hours may bad for UX.

Issue #2468, opened 2022-03-16
[Server] - sendVerificationCode mutation keeps returning FORBIDDEN


devformatters:
report
- [slack](https://mattersnews.slack.com/archives/CH8BLCMTN/p1647439407684179?thread_ts=1647404613.499369&cid=CH8BLCMTN)
- [slack](https://mattersnews.slack.com/archives/GA0R09FHN/p1647441618048589)

devformatters:
Don't forget to attach PR 🙌🏻

tx0c:
#2488 and #2478

PR #2474, opened 2022-03-17
fix(): skip recaptcha checking bot traffic

tx0c:
it has triggered many good gmail registration requests flagged as bot, need logs to know why

### DIFF
--- a/src/connectors/gcp/index.ts
+++ b/src/connectors/gcp/index.ts
@@ -101,11 +101,11 @@ class GCP {
 
     // fail for less than 0.5
     if (score < 0.5) {
-      return false
+      console.log("very likely bot traffic:", data)
     }
 
     // pass
-    return true
+    return score > 0.0
   }
 }
 


src/connectors/gcp/index.ts
```diff
@@ -101,11 +101,11 @@ class GCP {
 
     // fail for less than 0.5
     if (score < 0.5) {
-      return false
+      console.log("very likely bot traffic:", data)
```
guoliu, line 104: Shall we use `[logger](https://github.com/thematters/matters-server/blob/develop/src/common/logger.ts#L58)` here? If sentry is still accessable.

guoliu (approved):

src/connectors/gcp/index.ts
tx0c: I'm aware of that, but:
1. sentry is logging only warn, error, above severe logs
2. the logger.info can only log a string, not any extra parameters, or will have to use some thing like `JSON.stringify(data)` to include data; for simple logging, it's always easier to call console.log; let AWS infra configure something to monitor console logs to automatically save to CloudWatch might be a a better way

tx0c:
#2468

src/connectors/gcp/index.ts
guoliu: Sounds good, let's merge it in!

Issue #2460, opened 2022-03-08
[Server] - Return a specific error for wallet-registered users trying to login via email


tx0c:
should say some error like `wallet-registered users can only login via wallet`

https://matters.news/@MrLight/251685-被liker搞到自閉-bafyreigmziecksodbzatsloct5sze6a5st4pcfnvirfygyhatrl4ruph4y#Q29tbWVudDoxMDkxOTM5-Q29tbWVudDoxMDkyNTI4

devformatters:
report
- [slack](https://mattersnews.slack.com/archives/GA0R09FHN/p1646604811978619)

PR #2486, opened 2022-03-22
fix: error type for login user without password

Carolusian:
fix https://github.com/thematters/matters-server/issues/2460


Carolusian:
fix #2460

tx0c (approved):
LGTM

PR #2527, opened 2022-04-25
fix(omit-set-tags): change resetTags to omitTags at the beginning

tx0c:
use a small helper `sanitizeTags` to handle all cases;
- do not update tags if not provided in parameters


guoliu (approved):
Thanks for adding tests 🍀

PR #2621, opened 2022-07-14
feat(pin-tag): add action pin into action_tag

tx0c:
also changed some to TagSearchResultConnection


robertu7 (approved):

tx0c:
seems forgot to checkin the db/migrations file, add in #2629

PR #2622, opened 2022-07-15
fix(tag): temporary remove the clause of ON CONFLICT (content) DO ...

tx0c:
a better fix is to update it like in 6534e906ad3

but need another knex upgrade to have this new form:

    - insert(..).onConflict(knex.raw(...))

https://knexjs.org/guide/query-builder.html#onconflict


tx0c:
turn off some iscn related excessive logging by the way, deploy together to better track the 10% rate of iscn failures

PR #2646, opened 2022-07-27
feat(downvote): downvote counts return 0

Jasmine-liang:
Return 0 when querying the downvote counts


guoliu:
Can we add a deprecation directive so clients know to drop this query? One example [here](https://github.com/thematters/matters-server/blob/develop/src/types/notice.ts#L218), and more on directives [here](https://www.apollographql.com/docs/apollo-server/schema/directives/#default-directives).

robertu7 (approved):
LGTM.

Same comment with @guoliu 's.

Jasmine-liang:
Already added deprecation directive to downvotes.

PR #2649, opened 2022-07-28
fix(circle-notice): change .circle_new_broadcast back to target comment

tx0c:


robertu7 (approved):
Haven't follow-up the circle notification, but please go and test!

PR #2655, opened 2022-07-30
Update cors.ts for MattersProtocol.io

guoliu:
As title.


robertu7:
I guess we could close this PR since our domain already migrated to Cloudflare and got the same protection.

PR #2657, opened 2022-08-02
fix: returning('id') does not return integers in knex v2.1

### LINKED ISSUE #2647: [Server] Server fails to process `invoice.payment_succeeded` Stripe event
#### Tasks
- [ ] Investigate and fix Stripe events were not handled properly

#### References
- [slack](https://mattersnews.slack.com/archives/G01KESNNH8D/p1658923201776629)

Carolusian:
fix #2647

### DIFF
--- a/src/connectors/paymentService.ts
+++ b/src/connectors/paymentService.ts
@@ -541,7 +541,7 @@ export class PaymentService extends BaseService {
     const trx = await this.knex.transaction()
     try {
       // create subscription top up transaction
-      const transactionId = await trx('transaction')
+      const [trans] = await trx('transaction')
         .insert({
           amount,
           currency,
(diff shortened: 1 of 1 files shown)


src/connectors/paymentService.ts
```diff
@@ -556,18 +556,16 @@ export class PaymentService extends BaseService {
           targetType: undefined,
           targetId: undefined,
         })
-        .returning('id')
+        .returning('*')
```
tx0c, line 559: this line might not be necessary? I see you'll need only `trans.id`

tx0c (commented):
link to this line https://knexjs.org/changelog.html#_1-0-0-16-january-2022 in comments could be good, there might be other Breaking Changes as well:

```
Changed data structure from RETURNING operation to be consistent with SELECT;

robertu7 (approved):

robertu7:
wen to merge and test?

src/connectors/paymentService.ts
Carolusian: Updated to `.return('*')` because it has better compatibility, and is more consistent with the code in other places. For other code using `.return('*')`, the function is not breaking after being upgraded to knex v2.1.

PR #2659, opened 2022-08-03
chore: boost to thespace

Carolusian:


tx0c:
is this view a brand new one from scratch, or does it have a comparison to an existing

Carolusian:
> 

@tx0c similar to this one: https://github.com/thematters/matters-server/blob/develop/db/migrations/20220621013529_update_hottest_feed_boost_list.js

Just added `thespace` to matty user list

devformatters (approved):
Go. 🐧

PR #2660, opened 2022-08-03
add TAGS_RECOMMENDED_LIMIT in totalCount to related tags

tx0c:


robertu7 (approved):
is this to reduce response time?

PR #2676, opened 2022-08-10
feat(superlike-iscn): superlike with short url encoding

tx0c:
with iscn_id if available:
https://api.docs.like.co/#35e55922-559d-4bda-974f-5465fe023f48


tx0c:
closed in 5ae85868e522 in #2677

PR #2693, opened 2022-08-19
feat(i18n): store article translation to db instead of cache in redis

robertu7:


guoliu (approved):

src/queries/article/translation.ts
```diff
@@ -23,12 +39,30 @@ const resolver: ArticleToTranslationResolver = async (
...
+      title,
+      content,
+      summary,
+      language: target,
+    }
+    await atomService.upsert({
+      table: 'article_translation',
+      where: { articleId },
+      create: data,
+      update: { ...data, updatedAt: new Date() },
```
tx0c, line 55: updatedAt: this.knex.fn.now(),

src/queries/article/translation.ts
```diff
@@ -23,12 +39,30 @@ const resolver: ArticleToTranslationResolver = async (
     )
```
tx0c, line 39: does GCP charge by count or by text length? and do we translate the tags, and fix the translation length limit issue; `四萬字長文` seems always fail

although can be in a separate EPIC

robertu7: > does GCP charge by count or by text length? and do we translate the tags

[count by characters](https://cloud.google.com/translate/pricing), seems good to translate tags as well.


> and fix the translation length limit issue; 四萬字長文 seems always fail

“The Cloud Translation API is optimized for translating smaller requests. The recommended maximum length for each request is 5K characters ([code points](https://wikipedia.org/wiki/UTF-8)). However, the more characters that you include, the higher the response latency. For Cloud Translation - Advanced, the maximum number of code points for a single request is 30K. Cloud Translation - Basic has a maximum request size of 100K bytes. An alternative option is to use batch translations to send more content per request as a long running operation.”
https://cloud.google.com/translate/quotas#content

We can use `batchTranslateText`, will fix in this PR.

robertu7: `batchTranslateText` is more complex to call, shall take another time to fix [it](https://github.com/thematters/matters-server/issues/2694);

tx0c (approved):
looks big enough to merge; don't try to solve too many things in one PR

PR #2702, opened 2022-08-23
fix(ipfs-client): update server to be able use latest go-ipfs

### LINKED ISSUE #2288: [Server] Update server to be able use latest go-ipfs
### Tasks
- https://github.com/ipfs/go-ipfs/releases/tag/v0.10.0 has ipfs dag related breaking changes, caused our test/deploy pipeline fail and a temporary fix thematters/matters-server#2287 stick it to last good `v0.9.1`; but in the longer range we need to update it to latest

- need to research if npm package `ipfs-http-client` need a version bump

tx0c:
resolves #2288

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "graphql-schema-typescript": "^1.5.2",
     "graphql-upload": "^12.0.0",
     "helmet": "^4.6.0",
-    "ipfs-http-client": "^50.1.2",
+    "ipfs-http-client": "^56.0.3",
     "js-base64": "^3.7.2",
     "jsonwebtoken": "^8.5.1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -80,7 +80,7 @@
     "graphql-schema-typescript": "^1.5.2",
     "graphql-upload": "^12.0.0",
     "helmet": "^4.6.0",
-    "ipfs-http-client": "^50.1.2",
+    "ipfs-http-client": "^56.0.3",
```
tx0c, line 83: not the latest `^57.xx.yy` which is ESM only module, seems not compatible with current typescript transpling to target commonjs

PR #2709, opened 2022-08-30
feat(circle-comments-anchor): add comments filter to circle comments …

tx0c:
…(broadcast/discussions)

for thematters/matters-web#2683
thematters/matters-web#1903


robertu7 (approved):
Looks it's the same implementation as `src/queries/comment/article/comments.ts`, GO!

Issue #2631, opened 2022-07-19
[Server] RSS feed of user articles


devformatters:
#### Tasks
- [ ] Add API for RSS feed of user articles
- [ ] Require documentation

guoliu:
* Documentation on content feed format: https://www.notion.so/matterslab/Feed-format-8bafb64754de4434a1f3caea7ee5e4a9
* Documentation on publication flow / architecture: https://www.notion.so/matterslab/Publish-content-feed-to-IPNS-from-Matters-server-e294ee05fc76479c893dd6c24370caaf

guoliu:
Blocked by: https://github.com/thematters/matters-server/issues/2288

PR #2731, opened 2022-09-07
feat(ipns-infra): IPNS publishing

tx0c:
partI for #2631


src/connectors/ipfs/index.ts
```diff
@@ -1,17 +1,53 @@
...
+    const formData = new FormData()
+    const url = new URL(`${ipfsServerUrl}/key/import`)
+    url.searchParams.set('arg', name)
+    url.searchParams.set('format', 'pem-pkcs8-cleartext')
+    formData.append('file', Readable.from([pem]), 'keyfile')
+    const res = await fetch(url, {
+      method: 'POST',
+      body: formData,
+    })
+    console.log(new Date(), 'key/import res:', res)
```
robertu7, line 42: Is there any sensitive data that needs to be masked?

robertu7 (approved):
LGTM!

src/connectors/ipfs/index.ts
tx0c: no; the sample looks like this:

```js
2022-09-12T18:42:59.223Z key/import fetch res: {
  Name: 'ed-gen-test23-4-1663008179217',
  Id: 'k51qzi5uqu5dhdb6i7umgq6xcu5wt4appzz1c1kyqgh8folsecogs56j11if8n'
}
```

src/connectors/ipfs/index.ts
```diff
@@ -1,17 +1,53 @@
...
+    const imported = await res.json()
+    console.log(new Date(), 'key/import res json:', imported)
+    // return res.json()
+    return imported
```
tx0c, line 46: there is `ipfs.key.import` method, accept password
https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/KEY.md#ipfskeyimportname-pem-password-options

but the problem is only for js-ipfs implementation, different than go-ipfs

https://github.com/ipfs/js-ipfs/issues/3547#issuecomment-781991722

src/connectors/ipfs/index.ts
tx0c: many console log lines for debugging in the beginning days when running on dev; but eventually (in 2 ~ 4 weeks range) will be cleaned up

PR #2737, opened 2022-09-08
fix(dup-drafts): de-dup article-id from drafts

tx0c:


src/queries/article/user/articles.ts
```diff
@@ -27,12 +29,14 @@ const resolver: UserToArticlesResolver = async (
     inRangeEnd: input.filter?.inRangeEnd,
   })
 
-  return connectionFromPromisedArray(
-    draftService.dataloader.loadMany(
+  const drafts = uniqBy(
```
robertu7, line 32: It won't work if duplicated articles are across two pages?

src/queries/user/recommendation/newest.ts
```diff
@@ -44,10 +44,10 @@ export const newest: RecommendationToNewestResolver = async (
     .as('newest')
 
   const [_countRecord, articles] = await Promise.all([
-    MAX_ITEM_COUNT, // knex.select().from(query.clone().limit(MAX_ITEM_COUNT)).count().first(),
+    MAX_ITEM_COUNT, // always 500 // knex.select().from(query.clone().limit(MAX_ITEM_COUNT)).count().first(),
     knex
       .select()
-      .from(query.clone().limit(MAX_ITEM_COUNT))
```
robertu7, line 50: Remember why we used `clone' here?

robertu7 (approved):

src/queries/article/user/articles.ts
tx0c: still need `.distinctOn`

src/queries/user/recommendation/newest.ts
tx0c: the common pattern of

```js
const [ articlesCount, articles ] = Promise.all([
  500, // query.count(),
  query.clone().orderBy(...)...
])
```

no longer need the `.clone()` when the other clause is turning to const 500;

Issue #2732, opened 2022-09-07
[Server] incorrect processing of invoice payment webhook event


robertu7:
https://mattersnews.slack.com/archives/G01KESNNH8D/p1661996836710509

![Image](https://user-images.githubusercontent.com/4065233/188770028-b0cb5d2d-4e19-46ac-9820-708abe72c7f0.png)

PR #2756, opened 2022-09-18
fix: skip duplicated invoice.payment_succeeded webhook calls

Carolusian:
fix https://github.com/thematters/matters-server/issues/2732


src/connectors/paymentService.ts
```diff
@@ -619,6 +619,28 @@ export class PaymentService extends BaseService {
...
+  }: {
+    providerTxId: string
+    amount: number
+  }) => {
+    const splitTxs = await this.knex('transaction')
+      .select()
+      .where({
+        purpose: TRANSACTION_PURPOSE.subscriptionSplit,
+        state: TRANSACTION_STATE.succeeded,
+        remark: `stripe:${providerTxId}`,
```
robertu7, line 634: shall we add an index for this query?

robertu7 (approved):
LGTM

src/connectors/paymentService.ts
Carolusian: > shall we add an index for this query?

Or, shall we have a new field to refer to the parent transaction id?

PR #2863, opened 2022-10-17
refactor: use parentId instead of remark for tx split checking

Carolusian:


robertu7 (approved):
LGTM. Hot release?

Carolusian:
@robertu7 I guess can deploy together with next release as the existing check still works

PR #2866, opened 2022-10-17
fix(announcement): delete announcement translation

Jasmine-liang:
Delete announcement should delete its corresponding translation first.

### DIFF
--- a/src/mutations/system/deleteAnnouncements.ts
+++ b/src/mutations/system/deleteAnnouncements.ts
@@ -15,17 +15,15 @@ const resolver: MutationToDeleteAnnouncementsResolver = async (
 
   const itemIds = ids.map((id) => fromGlobalId(id).id)
 
-  await atomService
-    .deleteMany({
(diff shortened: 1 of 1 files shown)


src/mutations/system/deleteAnnouncements.ts
```diff
@@ -15,10 +15,17 @@ const resolver: MutationToDeleteAnnouncementsResolver = async (
...
   const itemIds = ids.map((id) => fromGlobalId(id).id)
 
-  await atomService.deleteMany({
-    table,
-    whereIn: ['id', itemIds],
-  })
+  await atomService
+    .deleteMany({
+      table: 'announcement_translation',
+      whereIn: ['id', itemIds],
```
tx0c, line 21: here should be `announcement_id`? from https://github.com/thematters/matters-server/blob/develop/db/migrations/20221009151817_create_announcement_translation.js#L9

the `announcement_translation` table has id but not the same one;

PR #2871, opened 2022-10-18
add queuejob of  polling USDT transtaction state 

gary02:


db/migrations/20221008153616_create_blockchain_transaction_table.js
```diff
@@ -13,7 +13,7 @@ exports.up = async (knex) => {
       .notNullable()
       .defaultTo('pending')
     t.timestamp('created_at').defaultTo(knex.fn.now())
-    t.timestamp('udpated_at').defaultTo(knex.fn.now())
+    t.timestamp('updated_at').defaultTo(knex.fn.now())
```
robertu7, line 16: Since the previous migration file was run, we should create another migration to alter this field.

src/common/environment.ts
```diff
@@ -99,8 +99,16 @@ export const environment = {
   logbookClaimerPrivateKey:
     process.env.MATTERS_LOGBOOK_CLAIMER_PRIVATE_KEY || '',
   alchemyApiKey: process.env.MATTERS_ALCHEMY_API_KEY || '',
+  curationContractAddress:
+    process.env.MATTERS_CURATION_CONTRACT_ADDRESS ||
+    (isProd ? '' : '0xa219C6722008aa22828B31A13ab9Ba93bB91222c'),
```
robertu7, line 104: why hardcode?

src/common/environment.ts
```diff
@@ -99,8 +99,16 @@ export const environment = {
...
 }
 
+export const USDTContractAddress = isProd
```
robertu7, line 107: same as above, use env instead?

src/common/environment.ts
```diff
@@ -99,8 +99,16 @@ export const environment = {
...
+  ? ''
+  : '0xfe4F5145f6e09952a5ba9e956ED0C25e3Fa4c7F1'
+export const USDTContractDecimals = isProd ? 6 : 18
```
robertu7, line 110: we can hardcode `18` for both dev and prod envs.

src/connectors/queue/payTo/payToByBlockchain.ts
```diff
@@ -0,0 +1,364 @@
...
+          const event = iface.parseLog(log)
+          if (
+            event.args.curator!.toLowerCase() ===
+              curatorAddress!.toLowerCase() &&
+            event.args.creator!.toLowerCase() ===
+              creatorAddress!.toLowerCase() &&
+            event.args.token!.toLowerCase() === tokenAddress.toLowerCase() &&
+            event.args.amount!.toString() ===
+              toTokenBaseUnit(amount, decimals) &&
+            (event.args.uri!.includes(cid) ||
```
robertu7, line 178: `uri` will be `ipfs://{CID}`, so here we have to strip out the prefix `ipfs://`?

src/connectors/queue/payTo/payToByBlockchain.ts
```diff
@@ -0,0 +1,364 @@
...
+
+    const blockchainTx =
+      await this.paymentService.findBlockchainTransactionById(tx.providerTxId)
+
+    if (!blockchainTx) {
+      throw new PaymentQueueJobDataError('blockchain transaction not found')
+    }
+
+    const provider = getProvider()
+    const confirms = 1
```
robertu7, line 214: 128 is a safe number to avoid reorg for Polygon?

src/mutations/user/payTo.ts
```diff
@@ -220,6 +220,7 @@ const resolver: MutationToPayToResolver = async (
           currency: PAYMENT_CURRENCY.USDT,
           purpose: TRANSACTION_PURPOSE[purpose],
         })
+      payToByBlockchainQueue.payTo({ txId: transaction.id })
```
robertu7, line 223: Since this queue job needs to [wait](https://github.com/thematters/matters-server/pull/2871/files#diff-345232780656783e4918d3a2f53a196e3e2b2998a8cad6f0e152d5cf39bc11d8R218) for the transaction to be confirmed, it's better to [delay](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#:~:text=delay%3A%20number%3B%20//%20An%20amount%20of%20milliseconds%20to%20wait%20until%20this%20job%20can%20be%20processed.%20Note%20that%20for%20accurate%20delays%2C%20both) (128 blocks / aprox. 5 mins) the job run when it's being added to queue, otherwise, there will be lots of jobs stuck in the queue.

robertu7 (changes requested):

db/migrations/20221008153616_create_blockchain_transaction_table.js
gary02: not run in  prod.  i have alter the schema in dev db by hand (This is not good..)

src/common/environment.ts
gary02: USDTContract address may be fixed?

src/connectors/queue/payTo/payToByBlockchain.ts
gary02: if some third parties call contract directly（and call matters graphQL api at the same time. Oh my god ）， we can't make sure they follow the same rule (it could be a gateway url).  

here may be we should validate the uri schema first

src/connectors/queue/payTo/payToByBlockchain.ts
gary02: the queue job is for updating transaction  state timely

src/common/environment.ts
gary02: exclude those private envs like privateKey...

db/migrations/20221008153616_create_blockchain_transaction_table.js
gary02: emm.. that is very bad,  i forget there not only dev db, but also local db,  i will fix it

src/common/environment.ts
robertu7: can put them in `.env.example`?

src/common/environment.ts
robertu7: hmm, i see, that is weird, seems not follow the convention..

src/connectors/queue/payTo/payToByBlockchain.ts
robertu7: we can check if `uri` has the `ipfs://`. If true, check if it's a valid CID (i think this check be omit) and match with the db record.

robertu7:
```ts
if ((/^ipfs:\/\//.test(uri))) {
   cid = uri.replace('ipfs://','')
}
```

PR #2891, opened 2022-10-28
feat(likecoin): put isCivicLiker checker into the queue

robertu7:


Jasmine-liang (approved):

src/connectors/queue/likecoin.ts
```diff
@@ -43,12 +54,20 @@ class LikeCoinQueue extends BaseQueue {
     })
   }
 
+  getCivicLiker = (data: GetCivicLikerData) => {
+    return this.q.add(QUEUE_JOB.getCivicLiker, data, {
+      priority: QUEUE_PRIORITY.NORMAL,
+      attempts: 1,
+    })
```
tx0c, line 61: for a single call, may save a return?

```js
getCivicLiker = (data: GetCivicLikerData) =>
  this.q.add(QUEUE_JOB.getCivicLiker, data, {
  ...
```

src/connectors/queue/likecoin.ts
```diff
@@ -94,6 +113,32 @@ class LikeCoinQueue extends BaseQueue {
...
+        keys: { id: likerId },
+        data: isCivicLiker,
+        expire: CACHE_TTL.LONG,
+      })
+
+      job.progress(100)
+      done(null, { likerId, isCivicLiker })
+    } catch (e) {
+      done(e)
+    }
```
tx0c, line 140: the `try -catch` may not be necessary, I see the bull supports Promise and throw err,

the throw exception is taken care of automatically

tx0c (approved):

PR #2901, opened 2022-10-31
add support settings

gary02:


src/queries/article/index.ts
```diff
@@ -119,6 +121,8 @@ export default {
     revisionCount,
     license: ({ license }: { license?: GQLArticleLicenseType }) =>
       license || ARTICLE_LICENSE_TYPE.cc_by_nc_nd_2,
+    supportRequest,
```
robertu7, line 124: Can omit resolvers of `supportRequest` and `supportReply` if the return value is inherited from database value directly.

robertu7 (changes requested):

src/queries/article/index.ts
gary02: For `supportReply` ,  some validation code will be added to check viewer if a donator in another PR.   And I hesitate to  add extra code to  handle native article problem here.

robertu7 (approved):

Jasmine-liang (approved):

PR #2917, opened 2022-11-06
pass support feedback setting to donated email template

gary02:


src/routes/pay/likecoin.ts
```diff
@@ -286,6 +290,7 @@ likecoinRouter.post('/', async (req, res, next) => {
         displayName: author.displayName,
         userName: author.userName,
       },
+      replyToDonator: draft.replyToDonator,
```
robertu7, line 293: This should be a boolean if [email](https://github.com/thematters/matters-email/pull/171/files#diff-05e59776f6f0ad1fb0e23a7340172f31a91a226c41ee17e66d9a52a3eb34611cR113) doesn't use the content of `replyToDonator`, otherwise, please confirm if empty string `''` is work as expected.

robertu7 (changes requested):

src/routes/pay/likecoin.ts
gary02: fixed

src/connectors/paymentService.ts
```diff
@@ -1072,4 +1078,102 @@ export class PaymentService extends BaseService {
...
+    const author = await atomService.findFirst({
+      table: 'user',
+      where: { id: article.authorId },
+    })
+    const draft = await atomService.findFirst({
+      table: 'draft',
+      where: { id: article.draftId },
+    })
+
+    const hasReplyToDonator =
```
robertu7, line 1118: `const hasReplyToDonator = !!draft.replyToDonator` is identical.

src/connectors/paymentService.ts
```diff
@@ -1072,4 +1078,102 @@ export class PaymentService extends BaseService {
...
+    }
+
+    notificationService.mail.sendPayment({
+      to: sender.email,
+      recipient: {
+        displayName: sender.displayName,
+        userName: sender.userName,
+      },
+      type: 'donated',
+      article: _article,
```
robertu7, line 1141: can simply like below, then `_article` is unnecessary.

```tsx
article: { ...article, id: tx.targetId, hasReplyToDonator }
```

robertu7 (approved):

src/connectors/paymentService.ts
gary02: is `!!draft.replyToDonator` clear in the case of  empty string?

robertu7: Yeah it would be clear enough (for JS developer), lol.

src/connectors/paymentService.ts
gary02: not all article fields （especially content field） need to pass to sendGrid api?

src/connectors/paymentService.ts
gary02: cool... I will fix it

PR #2935, opened 2022-11-14
add currency convertor related api

gary02:
- impl currency preference setting api
- mock currency exchange rates api

### DIFF
--- a/src/queries/user/index.ts
+++ b/src/queries/user/index.ts
@@ -1,4 +1,5 @@
 import { NODE_TYPES } from 'common/enums'
+import { AuthenticationError } from 'common/errors'
 import { toGlobalId } from 'common/utils'
 import {
   GQLAppreciationTypeResolver,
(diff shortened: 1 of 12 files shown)


src/queries/user/index.ts
```diff
@@ -109,7 +111,12 @@ const user: {
     info: (root) => root,
     // ipnsAddress,
     wallet: (root) => root,
-    settings: (root) => root,
+    settings: (root) => {
+      if (!root.id) {
```
robertu7, line 115: seems unnecessary since it's already protected by `settings: UserSettings! @auth(mode:"${AUTH_MODE.oauth}")`?

robertu7 (approved):

src/queries/user/index.ts
gary02: `oauth` allow visitors querying protected fields on `viewer`.

tx0c (approved):

PR #2956, opened 2022-11-18
Add currency conversion api

gary02:


src/common/enums/queue.ts
```diff
@@ -70,6 +70,10 @@ export const QUEUE_JOB = {
 
   // Circle
   transferTrialEndSubscriptions: 'transferTrialEndSubscriptions',
+
+  // ExchangeRate
+  updateFiat: 'updateFait',
```
robertu7, line 75: typo

src/common/enums/queue.ts
```diff
@@ -70,6 +70,10 @@ export const QUEUE_JOB = {
...
+  updateToken: 'updateToken',
```
robertu7, line 76: ..

.env.example
```diff
@@ -62,3 +62,4 @@ MATTERS_ALCHEMY_API_KEY=
 MATTERS_OPENSEA_API_KEY=
 MATTERS_POLYGON_CURATION_CONTRACT_ADDRESS=0xa219C6722008aa22828B31A13ab9Ba93bB91222c
 MATTERS_POLYGON_CURATION_CONTRACT_BLOCKNUM=28675517
+MATTERS_EXCHANGE_RATES_DATA_API_KEY=QLqGNNhnz8MHpwOh9AqTJv069po2aR09
```
robertu7, line 65: We are using Open Exchange Rates API as HKD to USD conversion for payout, is there any way to combine it with this? 

https://github.com/thematters/matters-server/blob/develop/src/connectors/paymentService.ts#L623-L629

src/connectors/exchangeRate/index.ts
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  private requestCoingeckoAPI = async (
+    bases: TokenCurrency[],
+    quotes: GQLQuoteCurrency[]
+  ): Promise<any | never> => {
+    const ids = bases.map((i) => TOKEN_TO_COINGECKO_ID[i]).join()
+    const vs_currencies = quotes.join()
+    try {
+      const reps = await axios.get(COINGECKO_API_URL, {
```
robertu7, line 197: There is no API key required from CoinGecko?

.env.example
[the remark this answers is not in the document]
robertu7: Besides production, shall we create another key for dev/local env?

src/connectors/__test__/exchangeRate.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    rate: 0.127826,
+    updatedAt: new Date('2022-11-18T06:28:03.000Z'),
+  },
+]
+
+describe('exchangeRate', () => {
+  const exchangeRate = new ExchangeRate()
+  beforeEach(() => {
+    // mock
+    exchangeRate.expire = 3
```
robertu7, line 93: 3 secs?

src/connectors/exchangeRate/index.ts
```diff
@@ -0,0 +1,251 @@
...
+]
+
+const TOKEN_TO_COINGECKO_ID = {
+  LIKE: 'likecoin',
+  USDT: 'tether',
+} as const
+
+const COINGECKO_API_URL = 'https://api.coingecko.com/api/v3/simple/price'
+const EXCHANGE_RATES_DATA_API_URL =
+  'https://api.apilayer.com/exchangerates_data/latest'
```
robertu7, line 44: No API versioning? It seems easy to break if any changes on API Layer.

src/connectors/exchangeRate/index.ts
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  updateFiatRates = async () => {
+    await this.updateRatesToCache(await this.fetchFiatRates())
+  }
+
+  private getRate = async (pair: Pair): Promise<GQLExchangeRate | never> => {
+    const data = (await this.cache.getObject({
+      keys: this.genCacheKeys(pair),
+      getter: async () => this.updateAllRatesAndGetRate(pair),
```
robertu7, line 83: Why don't just get the rate of this `pair` instead since there is concurrent issue if `getRates`'s `from` and `to` are `null` (means `getRate` of all pairs)?

robertu7 (changes requested):

.env.example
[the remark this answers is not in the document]
gary02: Not a fake API KEY .

> Besides production, shall we create another key for dev/local env?

Yes. another key for dev env is needed

src/common/enums/queue.ts
gary02: Oh..

.env.example
gary02: Open Exchange Rates API free plan only support USD related exchange rates querying and thus we should reuse `ExchangeRate.getRates` here

src/connectors/__test__/exchangeRate.test.ts
gary02: I mean to be 3 secs... will fix it

src/connectors/exchangeRate/index.ts
gary02: Yes
> Our Free API* has a rate limit of 10-50 calls/minute, and doesn't require API key
see https://www.coingecko.com/en/api/documentation

src/connectors/exchangeRate/index.ts
gary02: It seem that many exchange rates api do not use API versioning, like Open Exchange Rates API .  When those APIs change,  queue cron job will fail and send alert to Slack,  at the same time server still return old rates data in redis (TTL 10 days now). Hence,  I think we have enough time to response in the case of API breaking changing

src/connectors/__test__/exchangeRate.test.ts
gary02: Yes, I mean to be 3 secs... I will fix it

src/connectors/exchangeRate/index.ts
gary02: Yes, concurrent will be a issue. I will fix it

.env.example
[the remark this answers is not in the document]
robertu7: > Not a fake API KEY .

Then we should revoke it...?

robertu7: Yeah, let's see if we should change it in this PR.

src/connectors/__test__/exchangeRate.test.ts
```diff
@@ -0,0 +1,126 @@
...
+    rate: 0.127826,
+    updatedAt: new Date('2022-11-18T06:28:03.000Z'),
+  },
+]
+
+describe('exchangeRate', () => {
+  const exchangeRate = new ExchangeRate()
+  beforeEach(() => {
+    // mock
+    exchangeRate.expire = 3000 // 3 seconds
```
robertu7, line 93: Looks like I got it wrong, `expire` is second-basis, so `3000` is 3000 seconds...

src/connectors/exchangeRate/index.ts
```diff
@@ -0,0 +1,263 @@
...
+    if (isToken(from)) {
+      const data = await this.requestCoingeckoAPI([from], [to])
+      rate = this.parseCoingeckoData(data, { from, to })
+    } else if (isFiat(from)) {
+      const data = await this.requestExchangeRatesDataAPI(from, [to])
+      rate = this.parseExchangeRateData(data, { from, to })
+    } else {
+      throw new UnknownError('Unknown currency')
+    }
+    await this.updateRatesToCache([rate])
```
robertu7, line 122: the `getter` of `getObject` (L88) [already did the `storeObject`](https://github.com/thematters/matters-server/blob/develop/src/connectors/cache/index.ts#L105-L111), so here we can just fetch and return?

robertu7 (approved):
Overall LGTM, please take a look at the comments I left, thanks!

PR #2959, opened 2022-11-22
fix  DraftToArticleResolver resolver

gary02:


src/queries/article/index.ts
```diff
@@ -80,10 +81,15 @@ export default {
     topics: userTopics,
   },
   Article: {
-    // when the Article is duck'ing from a Draft it has articleId,
-    // otherwise when articleId is undefined, it's a native Article, just use 'id'
-    id: ({ articleId, id }: { articleId: string; id: string }) =>
-      toGlobalId({ type: NODE_TYPES.Article, id: articleId ?? id }),
+    id: ({ articleId, id }: { articleId: string; id: string }) => {
+      if (!articleId) {
+        logger.warn(
```
robertu7, line 86: still need to keep this logger?

robertu7 (approved):
Nice catch!

src/queries/article/index.ts
gary02: There is no guarantee that  wrong resolvers wll not be added to codebase  in future.

PR #2965, opened 2022-11-22
feat(ipfs-balancer): use a list of IPFS nodes instead, use all prior …

### LINKED ISSUE #2953: change matters-server to use a list of IPFS nodes instead, use all prior nodes as Secondary backup
current IPFS server usage is using addresses like `http://ipfs.dev.vpc:5001` trying use EC2 load balancer to use multiple IPFS nodes, but since IPNS feature requires using same node for same user's IPNS feed publishing, this transparent EC2 load balancer has been changed to use only 1 node, and hence no balancing actually,

Need to change this design to be handling a list of IPFS servers instead, pickup the first one as Primary active server, and in case the Primary IPFS server is problematic, retry with other servers

Suppose each IPFS server's availability is 99%, use multiple backups should be able improve whole system availability:

     1 - (1-99%) x (1-99%) x (1-99%) x (...) = 99.9999+%

because all past retired IPFS nodes were retired at 80~ 90% of disk usage, there are some extra 10 ~20GB disk space, should be okay to use all prior IPFS nodes as backup

or **in Application load balancing instead of EC2 load balancer**

tx0c:
…nodes as Secondary backup

resolves #2953

if one single IPFS node's Availability is 99%, use retry with multiple servers should improve HA to

    1 - (1-99%) x (1-99%) x (1-99%) x (...) = 99.9999+%


robertu7 (approved):
LGTM

PR #2992, opened 2022-12-01
fix(ipfs): use draft.articleId as article directory name

robertu7:


src/connectors/articleService.ts
```diff
@@ -316,7 +316,7 @@ export class ArticleService extends BaseService {
       })
       return
     }
-    const directoryName = `${kname}-with-${lastDraft.id}-${slugify(
+    const directoryName = `${kname}-with-${lastDraft.articleId}-${slugify(
```
tx0c, line 319: not only here, the fix will need to be in `publishedDrafts` all item's `id` will need to be `articleId`, and ipns-site-manager will need to change there?

robertu7: `@matters/ipns-site-generator` doesn't depend on our article or draft object from database, just render what we pass into it.

https://github.com/thematters/ipns-site-generator/blob/main/src/types.ts

src/connectors/feed/index.ts
```diff
@@ -94,9 +94,9 @@ export class Feed {
         date: draft.updatedAt,
         content: draft.content,
         tags: draft.tags || [],
-        uri: `./${draft.id}-${slugify(draft.title)}/`,
+        uri: `./${draft.articleId}-${slugify(draft.title)}/`,
```
tx0c, line 97: ok, then I made the actual change is here ^^

PR #3015, opened 2022-12-12
impl new version user & tag search

gary02:
- fix db setup script for tests with native database 
- preprocess searchV1 input (t2s, lowercase, strip spaces) 
- impl `UserService.searchV1`
- impl `TagService.SearchV1`

resolve https://github.com/thematters/matters-server/issues/2971

### DIFF
--- a/src/connectors/tagService.ts
+++ b/src/connectors/tagService.ts
@@ -670,7 +670,6 @@ export class TagService extends BaseService {
     }
   }
 
-  // TODO:
   searchV1 = async ({
     key,
     take,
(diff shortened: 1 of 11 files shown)


src/connectors/tagService.ts
```diff
@@ -685,7 +684,48 @@ export class TagService extends BaseService {
...
+    if (!_key) {
+      return { nodes: [], totalCount: 0 }
+    }
+
+    const mattyChoiceTagIds = environment.mattyChoiceTagId
+      ? [environment.mattyChoiceTagId]
+      : []
+
+    // search from original tags but return duplicate free tags
+    const queryIds = this.knex('search_index.tag')
```
robertu7, line 699: For performance reason, shall we `.limit` the query results?

src/connectors/userService.ts
```diff
@@ -600,7 +600,49 @@ export class UserService extends BaseService {
...
+    }
+
+    const baseQuery = this.knex('search_index.user')
+      .select('id')
+      .whereLike('display_name', `%${displayName}%`)
+      .orWhereLike('user_name', `%${userName}%`)
+      .orderByRaw('display_name = ? DESC', [displayName])
+      .orderByRaw('user_name = ? DESC', [userName])
+      .orderByRaw('display_name ~ ? DESC', [displayName])
+      .orderByRaw('num_followers DESC NULLS LAST')
```
robertu7, line 618: Same question as above, shall we `.limit` the query results?

robertu7 (approved):

src/connectors/tagService.ts
gary02: limit do be set in L723？

src/connectors/userService.ts
gary02: limit is set in L639

src/connectors/tagService.ts
robertu7: But the query performance of L702 impacted by `queryIds` right?

src/connectors/userService.ts
```diff
@@ -600,7 +600,49 @@ export class UserService extends BaseService {
...
+
+    if (!userName) {
+      return { nodes: [], totalCount: 0 }
+    }
+
+    const baseQuery = this.knex('search_index.user')
+      .select('id')
+      .whereLike('display_name', `%${displayName}%`)
+      .orWhereLike('user_name', `%${userName}%`)
+      .orderByRaw('display_name = ? DESC', [displayName])
```
tx0c, line 615: here need to check if search key startsWith('@'), then search on user_name only, no matching on `display_name`

src/connectors/userService.ts
```diff
@@ -600,7 +600,49 @@ export class UserService extends BaseService {
...
+      .orderByRaw('user_name = ? DESC', [userName])
+      .orderByRaw('display_name ~ ? DESC', [displayName])
```
tx0c, line 617: and here, if startsWith('@') then no check display_name

src/connectors/tagService.ts
gary02: Yes， but if `queryIds` is limited,  the final search result will lack some hit items?

PR #3018, opened 2022-12-12
fix(liker): set timeout for isCivicLiker query

robertu7:
Fix `isCivicLiker` query resolver and queue:
* Set a timeout (2000ms);
* Use unique id as queue job id;
* Remove job after complete or failed;

### DIFF
--- a/src/connectors/queue/likecoin.ts
+++ b/src/connectors/queue/likecoin.ts
@@ -146,7 +146,7 @@ class LikeCoinQueue extends BaseQueue {
       // update cache
       await cacheService.storeObject({
         keys: { id: likerId },
-        data: true,
+        data: isCivicLiker,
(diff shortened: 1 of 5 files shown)


src/connectors/queue/likecoin.ts
```diff
@@ -98,42 +111,53 @@ class LikeCoinQueue extends BaseQueue {
...
+
+      await invalidateFQC({
+        node: { type: NODE_TYPES.User, id: userId },
+        redis: this.cacheService.redis,
+      })
+
+      // update cache
+      await cacheService.storeObject({
+        keys: { id: likerId },
+        data: true,
```
gary02, line 149: data should be isCivicLiker?

gary02 (approved):

PR #3023, opened 2022-12-13
Record user last seen timestamp in database

gary02:


tx0c (approved):

tx0c:
remember there still need a place (cron job / lambda) to update for those users active less than 1 hour,
need the cron job to periodically check from Redis cache, and UPDATE INTO users table

PR #3047, opened 2022-12-20
feat(user-retention): create user retetion table

gary02:


db/migrations/20221212215759_user_retention_history_table.js
```diff
@@ -0,0 +1,17 @@
...
+
+const table = 'user_retention_history'
+
+exports.up = async (knex) => {
+  await knex('entity_type').insert({ table })
+  await knex.schema.createTable(table, (t) => {
+    t.bigIncrements('id').primary()
+    t.bigInteger('user_id').unsigned().notNullable()
+    t.enu('state', ['NEWUSER', 'ALERT', 'INACTIVE', 'NORMAL', 'ACTIVE'])
+    t.timestamp('created_at').defaultTo(knex.fn.now())
```
tx0c, line 11: should it also have `updated_at` ? for use when `state` changes

gary02: when `state` changes, a new state record with `created_at` will be inserted,   a `updated_at` seem to be duplicating?

PR #3065, opened 2023-01-06
migrate sendmail job to sqs/lambda

gary02:
1. refactor aws class and add `ArticleService.sendArticleFeedMsgToSQS`
2. remove `NotificationQueue`
3. move sendgrid logic to [sendmail lambda](https://github.com/thematters/lambda-handlers/blob/main/handlers/sendmail.ts)
4. connect `MailService` and sendmail lambda


src/connectors/aws/index.ts
```diff
@@ -196,17 +194,19 @@ export class AWSService {
...
-    MessageGroupId,
-    MessageBody,
+    messageBody,
+    queueUrl,
+    messageGroupId,
   }: {
-    MessageGroupId: string
-    MessageBody: any
+    messageBody: any
+    queueUrl: string
```
robertu7, line 202: how about it's an enum type instead of passing a URL directly?

robertu7 (approved):

src/connectors/aws/index.ts
gary02: good idea~

PR #3066, opened 2023-01-10
fix(article-timestamp): use original article's publishing timestamp i…

tx0c:
…nstead


src/connectors/articleService.ts
```diff
@@ -965,12 +973,12 @@ export class ArticleService extends BaseService {
     const { hits, ...rest } = await this.meili.index('articles').search(key, {
       limit: take,
       offset: skip,
-      filter: ['state = active'],
+      // filter: ['state = active'],
```
robertu7, line 976: this change doesn't relate to article publishing?

robertu7 (approved):

src/connectors/articleService.ts
tx0c: right; but the filter is for articles searchV1 and no longer needed, could be placed into another PR but...

PR #3071, opened 2023-01-10
migrate `likecoinQueue `  to SQS/lambda

gary02:


.env.example
```diff
@@ -13,6 +13,9 @@ MATTERS_AWS_CLOUD_FRONT_ENDPOINT=
 MATTERS_AWS_IPFS_ARTICLES_QUEUE_URL="https://sqs.ap-southeast-1.amazonaws.com/903380195283/my-queue.fifo"
 MATTERS_AWS_ARTICLES_SNS_TOPIC="arn:aws:sns:ap-southeast-1:903380195283:ipfs-articles-dev"
 MATTERS_AWS_MAIL_QUEUE_URL="https://sqs.ap-southeast-1.amazonaws.com/903380195283/mail-dev"
+MATTERS_AWS_LIKECOIN_LIKE_QUEUE_URL=""
```
robertu7, line 16: might need a better to organize these urls since EB has a [cap on env size](https://stackoverflow.com/questions/60279738/elastic-beanstalk-maximum-environment-variable-length)

robertu7 (approved):

PR #3074, opened 2023-01-11
fix(tagsearch): fix wrong ordering in tag search result

gary02:


src/connectors/tagService.ts
```diff
@@ -701,31 +701,17 @@ export class TagService extends BaseService {
...
       .andWhere((builder: Knex.QueryBuilder) => {
         builder.whereNotIn('id', mattyChoiceTagIds)
-      }) /* .modify((builder: Knex.QueryBuilder) => {
-        if (keyOriginal) {
-          builder.orderByRaw('content = ? DESC', [keyOriginal]) // always show original key before normalized exact match at first
-        }
-      }) */
+      })
       .orderByRaw('content = ? DESC', [_key]) // always show exact match at first
       .orderBy('num_articles', 'desc')
```
tx0c, line 716: can also do nulls last, then coalesce will not be necessary

    .orderByRaw('num_articles DESC NULLS LAST')

PR #3088, opened 2023-01-30
fix(article):   `Article` add `pending` `error` state  

gary02:
- add new states to  `Article`
- correct  article records state in database
- turn off most logging output in local testing


db/migrations/20230129101723_fix_article_state.js
```diff
@@ -0,0 +1,15 @@
+exports.up = async (knex) => {
+  await knex.raw(/*sql*/ `
+  UPDATE article SET state = 'error' WHERE id IN (
+    SELECT article.id FROM draft 
+    LEFT JOIN article ON article.draft_id = draft.id
+      WHERE draft.publish_state != 'published' AND  article.draft_id IS NOT NULL
```
robertu7, line 6: all drafts that `publish_state` in `unpublished`, `error`, and `pending` are included?

src/connectors/queue/__test__/publication.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+describe('publicationQueue.publishArticle', () => {
+  test('publish not pending draft', async () => {
+    const notPendingDraftId = '1'
+    const draft = await publicationQueue.draftService.baseFindById(
+      notPendingDraftId
+    )
+    expect(draft.state).not.toBe(PUBLISH_STATE.pending)
+
+    const job = await publicationQueue.publishArticle({
```
robertu7, line 14: aha, now I know how to test an async queue job!

src/connectors/queue/publication.ts
```diff
@@ -361,6 +337,11 @@ class PublicationQueue extends BaseQueue {
         )
       }
 
+      await this.articleService.baseUpdate(article.id, {
```
robertu7, line 340: shall we move [update `draft.publishState` to `published`](https://github.com/thematters/matters-server/blob/b95b0de1c1d79986a7d9ef1e72e6797f0c596a9b/src/connectors/queue/publication.ts#L141) to here?

robertu7 (approved):

db/migrations/20230129101723_fix_article_state.js
gary02: Yes, this is what the sql does.  You mean `draft.publish_state != 'published'` should be `draft.publish_state = 'error'` ?

PR #3100, opened 2023-02-06
ci(envs): let app read ENVS from AWS Systems Manager Parameter Store

gary02:
for more info, see
https://www.notion.so/matterslab/Elastic-Beanstalk-Read-envs-from-SSM-Parameter-Store-87acd49624dc41989edc479fa860190c


robertu7 (approved):
LGTM!

PR #3138, opened 2023-02-14
feat(exchange-rate): move exchange rate update cronjob to lambda

gary02:


src/connectors/exchangeRate/index.ts
```diff
@@ -151,31 +131,6 @@ export class ExchangeRate {
...
-    for (const f of fiatCurrencies) {
-      const data = await this.requestExchangeRatesDataAPI(f, quoteCurrencies)
-      for (const q of quoteCurrencies) {
-        rates.push(this.parseExchangeRateData(data, { from: f, to: q }))
-      }
-    }
-    return rates
-  }
-
   private parseCoingeckoData = (
```
robertu7, line 134: shall we still keep these methods?

robertu7 (approved):

src/connectors/exchangeRate/index.ts
gary02: In practice, we will not query apis  directly to get exchange rates  ,  as  exchange rates have store in cache for 10 days long. But considering the situation we setup new env  for matters server ( matters-server-stage, for example),  exchange rates will not exist in cache until cron job run its first time.

PR #3176, opened 2023-03-01
feat(readreplica): migrate heavy read queries to replia

gary02:


src/connectors/systemService.ts
```diff
@@ -214,7 +214,7 @@ export class SystemService extends BaseService {
    * Find the url of an asset by a given id.
    */
   findAssetUrl = async (id: string): Promise<string | null> => {
-    const result = await this.baseFindById(id, 'asset')
+    const result = await this.knexRO.from('asset').where({ id }).first()
```
robertu7, line 217: Would this affect by the replication lag? Like uploading images to user avatar or draft editor.

robertu7 (approved):

src/connectors/systemService.ts
gary02: oh, yes, it would be affect in these cases. I will revert this change

PR #3253, opened 2023-03-27
feat: connections per article update on `putDraft`

gary02:


src/mutations/draft/putDraft.ts
```diff
@@ -113,46 +115,37 @@ const resolver: MutationToPutDraftResolver = async (
...
-          userId: article.authorId,
-          targetId: viewer.id,
-        })
+  // handle Collection
+  const newCollectionIds =
+    collection == null
+      ? []
+      : _.uniq(
+          collection
+            .filter(_.isString)
```
robertu7, line 124: As the schema defined, `collection` is `[ID]`, an array of strings, so is this unnecessary?

robertu7 (approved):

src/mutations/draft/putDraft.ts
gary02: if `collection` is` [ID]` is `[ID!]`, it`s an array of strings.  Maybe the schema is wrong?

robertu7: Hmm, yes it should be `[ID!]`. Let's leave it here cuz any schema changes require client-side too.

PR #3270, opened 2023-04-06
[hotfix - master] fix superlike & de-dup totalWordCount

robertu7:


gary02 (approved):

src/connectors/articleService.ts
```diff
@@ -624,24 +624,37 @@ export class ArticleService extends BaseService {
...
+      table: 'article',
+      where: { draftId: targetArticle.draftId },
+    })
+
+    // update db
+    for (const article of articles) {
+      await this.baseUpdate(article.id, {
+        state: ARTICLE_STATE.archived,
+        sticky: false,
+        updatedAt: new Date(),
```
tx0c, line 642: in many cases, we have changed this to `updatedAt: knex.fn.now()` which has better timestamp precision, better in later analysis

tx0c: and this findMany & for-loop look like might be just one call of `await atomService.updateMany` ?

src/queries/user/articleCount.ts
```diff
@@ -4,13 +4,23 @@ import { UserStatusToArticleCountResolver } from 'definitions'
...
-    table: 'article',
-    where: { authorId: id, state: ARTICLE_STATE.active },
-  })
-  return count
+  const result = await knex('article as a1')
+    .where({ 'a1.author_id': id })
+    .whereNotExists(
+      knex('article as a2')
+        .select(1)
+        .where('a2.draft_id', '=', knex.raw('a1.draft_id'))
```
tx0c, line 14: .whereRaw('a2.draft_id = a1.draft_id')

tx0c (approved):

src/connectors/articleService.ts
robertu7: I will make another PR to `develop`

src/queries/user/articleCount.ts
robertu7: I will make another PR to `develop`

PR #3272, opened 2023-04-06
feat: improve code

robertu7:


src/connectors/atomService.ts
```diff
@@ -227,8 +235,24 @@ export class AtomService extends DataSource {
...
+    columns = '*',
+  }: UpdateManyInput) => {
+    const action = this.knex
+
+    if (where) {
+      action.where(where)
+    }
+    if (whereIn) {
+      action.whereIn(...whereIn)
+    }
```
tx0c, line 252: if no `where` nor `whereIn` this will become dangerous of changing all rows:

    UPDATE table SET data ...

PR #3273, opened 2023-04-07
feat: impl oss user restriction

### LINKED ISSUE #3243: Create APIs for OSS user management
### Tasks
- [x] DB changes
- [x] Create APIs for OSS user management
- [x] Test cases

### References
- [Product doc](https://www.notion.so/matterslab/64a78214148645c5a8a09185923dbac8)

### Blockers
- https://github.com/thematters/developer-resource/issues/204

gary02:
- add db migration
- update graphQL schema
-  impl related resolvers
- refactors

resolve #3243

### DIFF
--- a/db/migrations/20230406025628_create_user_restriction_table.js
+++ b/db/migrations/20230406025628_create_user_restriction_table.js
@@ -7,7 +7,7 @@ exports.up = async (knex) => {
   await knex.schema.createTable(table, (t) => {
     t.bigIncrements('id').primary()
     t.bigInteger('user_id').unsigned().notNullable()
-    t.enu('restriction', ['articleHottest', 'articleNewest']).notNullable()
+    t.enu('type', ['articleHottest', 'articleNewest']).notNullable()
     t.timestamp('created_at').defaultTo(knex.fn.now())
 
     t.index(['user_id'])
--- a/db/migrations/20230406025628_create_user_restriction_table.js
+++ b/db/migrations/20230406025628_create_user_restriction_table.js
@@ -1,0 +1,18 @@
+const { baseDown } = require('../utils')
+
+const table = 'user_restriction'
+
+exports.up = async (knex) => {
+  await knex('entity_type').insert({ table })
+  await knex.schema.createTable(table, (t) => {
+    t.bigIncrements('id').primary()
+    t.bigInteger('user_id').unsigned().notNullable()
+    t.enu('restriction', ['articleHottest', 'articleNewest']).notNullable()
+    t.timestamp('created_at').defaultTo(knex.fn.now())
+
+    t.index(['user_id'])
+    t.foreign('user_id').references('id').inTable('user')
+  })
+}
+
+exports.down = baseDown(table)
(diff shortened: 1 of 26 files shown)


db/migrations/20230406025628_create_user_restriction_table.js
```diff
@@ -0,0 +1,18 @@
...
+
+const table = 'user_restriction'
+
+exports.up = async (knex) => {
+  await knex('entity_type').insert({ table })
+  await knex.schema.createTable(table, (t) => {
+    t.bigIncrements('id').primary()
+    t.bigInteger('user_id').unsigned().notNullable()
+    t.enu('type', ['articleHottest', 'articleNewest']).notNullable()
+    t.timestamp('created_at').defaultTo(knex.fn.now())
```
robertu7, line 11: shall we need a `state` (`archived`, `active`) field so we don't need to delete the record?

src/connectors/userService.ts
```diff
@@ -2092,6 +2094,67 @@ export class UserService extends BaseService {
...
+    })
+  }
+
+  updateRestrictions = async (id: string, types: GQLUserRestrictionType[]) => {
+    const olds = (await this.findRestrictions(id)).map(({ type }) => type)
+    const news = [...new Set(types)]
+    const toAdd = news.filter((i) => !olds.includes(i))
+    const toDel = olds.filter((i) => !news.includes(i))
+    await Promise.all(
+      toAdd
```
robertu7, line 2118: can use `[...toAdd.map(...), ...toDel.map(...)]` instead

db/migrations/20230406025628_create_user_restriction_table.js
```diff
@@ -0,0 +1,18 @@
...
+
+    t.index(['user_id'])
```
robertu7, line 13: also index `type`?

robertu7 (approved):
🈂️

src/connectors/userService.ts
gary02: so neat!

db/migrations/20230406025628_create_user_restriction_table.js
gary02: enum column may not benefit too much from indexing ?

db/migrations/20230406025628_create_user_restriction_table.js
gary02: the `state` col will make db operation a little more complex, and according the production requirement by now,  we do not need keep the history of the restriction operation?

tx0c: some `state` kind of soft delete is still better; maybe some time later we will want to analyze

db/migrations/20230406025628_create_user_restriction_table.js
robertu7: we use filtering by `type` in [recommendation resolvers](https://github.com/thematters/matters-server/pull/3273/files#diff-f090852af5161630e96eabe822e33d6c3dc8aefcc8812c00807235f6a39a4728R45-R50), any performance impact?

gary02: Indexed enum column first scan index and read, let say 30%, rows row by row from disk,  comparing to batch read all rows from disk in the un-indexed case,   it hard to say which one can be faster.

PR #3286, opened 2023-04-18
feat(in-wallet): allow both .news & .town domain

tx0c:
use ${environment.siteDomain} to construct website link

### DIFF
--- a/src/mutations/user/generateSigningMessage.ts
+++ b/src/mutations/user/generateSigningMessage.ts
@@ -28,21 +28,21 @@ const resolver: MutationToGenerateSigningMessageResolver = async (
   const expiredAt = new Date(+createdAt + 10 * 60e3) // 10 minutes
 
   // create the message to be sign'ed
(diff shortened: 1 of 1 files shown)


robertu7 (approved):

src/mutations/user/generateSigningMessage.ts
```diff
@@ -28,21 +28,21 @@ const resolver: MutationToGenerateSigningMessageResolver = async (
   const expiredAt = new Date(+createdAt + 10 * 60e3) // 10 minutes
 
   // create the message to be sign'ed
-  const signingMessage = `${environment.siteDomain} wants you to sign in with your Ethereum account:
+  const signingMessage = `${environment.domain} wants you to sign in with your Ethereum account:
```
gary02, line 31: here would be `server.matters.town` ?

robertu7: based on envs on EB, yeah...

PR #3288, opened 2023-04-18
feat(in-wallet): allow both .news & .town domain

tx0c:
use ${environment.siteDomain} to construct website link


robertu7:
Still need to strip the protocol `https://`

Issue #3254, opened 2023-03-27
Dedup articles on publication


robertu7:
### Reports
* https://matters.canny.io/bug-reports/p/0fa964

### Tasks
* [x] Investigate and fix article publication

### References
- https://github.com/thematters/matters-server/pull/2737
- [Thread discuss](https://mattersnews.slack.com/archives/GA0R09FHN/p1679577077346339?thread_ts=1679570219.850109&cid=GA0R09FHN)

devformatters:
@robertu7 Could you provide more info about this issue. 🙏🏻

gary02:
related fix:  https://github.com/thematters/matters-server/pull/3267

PR #3293, opened 2023-04-19
fix(publication): specify job id to avoid duplicate articles

gary02:
fix: #3254


src/connectors/queue/publication.ts
```diff
@@ -46,6 +46,7 @@ class PublicationQueue extends BaseQueue {
       { draftId, iscnPublish },
       {
         priority: QUEUE_PRIORITY.CRITICAL,
+        jobId: `${QUEUE_JOB.publishArticle}:${draftId}`,
```
robertu7, line 49: nice, totally forgot there is a `jobId` parameter...

robertu7 (approved):

PR #3333, opened 2023-05-05
feat(search): deprecate search version in api and retire elasticsearch

gary02:
- deprecate search version
- remove es and meili related code


src/types/system.ts
```diff
@@ -234,8 +234,9 @@ export default /* GraphQL */ `
     record: Boolean
     oss: Boolean
 
-    "use the api version; omit to use latest version"
+    "deprecated, make no effect"
     version: SearchAPIVersion
```
robertu7, line 238: can use `@deprecated` directive?

robertu7 (changes requested):

src/types/system.ts
gary02: https://matterslab.slack.com/archives/C88CK7Q7L/p1669339562477699

robertu7 (approved):

robertu7 (approved):

PR #3340, opened 2023-05-09
fix(upload): strip envPrefix from cloudflare image key

gary02:
to keep compatibility with old keys in database


src/connectors/cloudflare/index.ts
```diff
@@ -82,10 +81,7 @@ export class CloudflareService {
 
     try {
       const resData = await res.json()
-
-      // assert "success": true,
-
-      return resData.result.id // the key
+      console.log(new Date(), 'CloudflareService upload image:', resData)
```
robertu7, line 84: shall we keep this logger?

robertu7 (approved):

PR #3386, opened 2023-05-31
feat(article): add article sensitive fields and mutations

robertu7:
* Add `Article.sensitiveByAuthor` and `Article.sensitiveByAdmin`;
* Add `updateArticleSensitive` for admin (OSS);
* Revise `editArticle` to support `sensitive` input;

### DIFF
--- a/src/mutations/article/updateArticleSensitive.ts
+++ b/src/mutations/article/updateArticleSensitive.ts
@@ -1,0 +1,20 @@
+import { fromGlobalId } from 'common/utils'
+import { MutationToUpdateArticleSensitiveResolver } from 'definitions'
+
+const resolver: MutationToUpdateArticleSensitiveResolver = async (
+  _,
+  { input: { id, sensitive } },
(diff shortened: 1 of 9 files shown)


src/mutations/article/updateArticleSensitive.ts
```diff
@@ -0,0 +1,20 @@
+import { fromGlobalId } from 'common/utils'
+import { MutationToUpdateArticleSensitiveResolver } from 'definitions'
+
+const resolver: MutationToUpdateArticleSensitiveResolver = async (
+  _,
+  { input: { id, sensitive } },
+  { viewer, dataSources: { articleService, draftService } }
```
gary02, line 7: `viewer` is not used？

robertu7: Yeah it's not in use since `@auth(mode: "${AUTH_MODE.admin}")` does the authentication, we can remove this line.

PR #3401, opened 2023-06-07
feat: update apollo-server to v4 and graphql deps

gary02:
- update apollo server from v3 to v4, including install [renamed packages](https://www.apollographql.com/docs/apollo-server/migration/#renamed-packages) and `keyv`
- update apollo construction code (like move `dataSources` to `context`) as apollo v4 [change its constructor options](https://www.apollographql.com/docs/apollo-server/migration/#removed-constructor-options)
- Errors now dirive from `graphql`'s `GraphQLError` instead of removed `ApolloError`
- update testClient as [`GraphQLResponse` object are refactored](https://www.apollographql.com/docs/apollo-server/migration/#graphqlresponse)
- update Opensea connector as `RESTDataSource` API changed
- CacheService use redis client instead of deprecated `redisCache`
- remove `requestContext`, use `Context` type directly
- update @thematters/apollo-cache-response to 2.0.0-alpha.0 and related codes(latest)
- update graphql-constraint-directive to 5.1.1 (latest)
- update @graphql-tools to 10.0.0 (latest)
- update graphql-upload to 13.0.0 to update its graphql.js to 16. 14.0.0+ is esm module
- remove graphql-depth-limit, as its type package do not use graphql.js 16
- update directives and related codes as graphql-tools deprecate `SchemaDirectiveVisitor` 
- @objectCache remove unused OBJECT location
- remove unused dev dependencys


src/mutations/article/putChapter.ts
```diff
@@ -1,9 +1,9 @@
-import { UserInputError } from 'apollo-server-core'
```
robertu7, line 1: nice catch...

robertu7 (approved):
🚀

PR #3462, opened 2023-07-14
fix(collection-articles): fix `reorderArticles` logic bug

gary02:


src/connectors/collectionService.ts
```diff
@@ -198,7 +198,9 @@ export class CollectionService extends BaseService {
     moves: Array<{ articleId: string; newPosition: number }>
   ) => {
     const [collectionArticles, count] =
-      await this.findAndCountArticlesInCollection(collectionId, {})
+      await this.findAndCountArticlesInCollection(collectionId)
+    console.dir(collectionArticles, { depth: null })
```
robertu7, line 202: forgot to remove?

robertu7 (approved):

src/connectors/collectionService.ts
gary02: right...

PR #3482, opened 2023-07-24
[hotfix - master] fix(limit): fix incorrect rate limit counting

robertu7:


src/types/directives/rateLimit.ts
```diff
@@ -40,15 +40,13 @@ const checkOperationLimit = async ({
   }
```
tx0c, line 41: use a local `const cutoff = current - period` can hold the cutoff timestamp and save a few ops?

and the `operationLog` is supposed to save timestamps in decreasing order, so some alternative way be clear and better?
1. a simple `array.findIndex` can find out first index of timestamp `< cutoff`,
2. or just use a for loop can `return false early` when count to `limit`?

PR #3516, opened 2023-08-06
feat(limit): change gql cost limit to 10k

robertu7:


robertu7:
https://github.com/thematters/matters-server/pull/3517

PR #3528, opened 2023-08-11
feat: update `setUserName`

gary02:
- set default displayName
- check userName existence before creating Articles/Collections/Tags/Comments/Circles


db/sql/create-table-search-index-user.sql
```diff
@@ -13,7 +13,7 @@ CREATE TABLE IF NOT EXISTS search_index.user AS
     FROM action_user
     GROUP BY 1
   ) t ON target_id=u.id
-  -- WHERE state IN ('active', 'onboarding')
```
robertu7, line 16: confirm to include non-active users?

robertu7 (approved):

PR #3529, opened 2023-08-11
feat: check email if verified before set password and fix typing

gary02:
- check email if verified before set password
- correct `user.email` type to be `string | null` and pass build

### DIFF
--- a/src/mutations/user/addCredit.ts
+++ b/src/mutations/user/addCredit.ts
@@ -8,10 +8,10 @@ import {
   TRANSACTION_PURPOSE,
 } from 'common/enums'
 import {
-  AuthenticationError,
   PaymentAmountInvalidError,
   PaymentAmountTooSmallError,
   ServerError,
+  ForbiddenError,
 } from 'common/errors'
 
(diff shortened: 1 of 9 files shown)


src/mutations/user/addCredit.ts
```diff
@@ -8,19 +8,19 @@ import {
...
 } from 'common/errors'
 
 const resolver: GQLMutationResolvers['addCredit'] = async (
   _,
   { input: { amount } },
   { viewer, dataSources: { atomService, paymentService } }
 ) => {
-  if (!viewer.id) {
-    throw new AuthenticationError('visitor has no permission')
+  if (!viewer.email) {
```
robertu7, line 22: email is not required by payment services (add credit, donation, etc.), right?

robertu7 (approved):

src/mutations/user/addCredit.ts
gary02: creating stripe customers need email info ?

gary02: https://matterslab.slack.com/archives/G8877EQMS/p1692866329164109

PR #3533, opened 2023-08-14
feat(onboarding): deprecate user onboarding state

robertu7:


db/sql/create-indexes.sql
```diff
@@ -59,7 +59,7 @@ CREATE TABLE IF NOT EXISTS public.article_read_time_materialized AS
     sum(article_read_count.read_time) AS sum_read_time
    FROM article_read_count
      JOIN "user" ON "user".id = article_read_count.user_id
-  WHERE article_read_count.user_id IS NOT NULL AND ("user".state = ANY (ARRAY['active'::text, 'onboarding'::text]))
```
robertu7, line 62: should also confirm if https://github.com/thematters/matters-server/blob/develop/db/migrations/20210726095042_create_article_read_time_view.js#L16 also need to be revised

gary02 (approved):

PR #3550, opened 2023-08-23
feat(move-ipns): tear down most of IPNS features

tx0c:
matters-server will still keep create & manage IPNS keys, not relying on 3rd party


gary02 (approved):

robertu7 (approved):

robertu7:
linked to https://github.com/thematters/lambda-handlers/issues/71

PR #3636, opened 2023-10-05
feat(audit-log): implement

gary02:
implement audit logging documented in https://www.notion.so/matterslab/Audit-Logs-9cdbede5196b4043962954cab108e893

related PR: https://github.com/thematters/lambda-handlers/pull/91


src/connectors/userService.ts
```diff
@@ -411,6 +411,12 @@ export class UserService extends BaseService {
           language: user.language,
         })
       }
+      auditLog({
+        actorId: userId,
+        action: 'update_email',
```
robertu7, line 416: enums for action?

src/connectors/userService.ts
```diff
@@ -434,6 +440,7 @@ export class UserService extends BaseService {
     if (!user.email || !user.emailVerified) {
       throw new ForbiddenError('email not verified')
     }
+    auditLog({ actorId: user.id, action: 'update_password' })
```
robertu7, line 443: For security and product analysis reasons, I believe we should not only log the successful result of an action but also the attempt and failure result:

```
auditLog({ actorId: user.id, action: 'update_password', status: 'pending' })
try {
  ...
  auditLog({ actorId: user.id, action: 'update_password', status: 'successed' })
} catch (error) {
  auditLog({ actorId: user.id, action: 'update_password', status: 'failed' })
  ...
}
```

robertu7 (changes requested):

src/common/logger.ts
```diff
@@ -63,6 +63,24 @@ export const getLogger = (name: string) => {
...
+export const auditLog = (data: {
+  actorId: string | null
+  action: string
+  entity?: TableName
+  entityId?: string
+  oldValue?: string | null
+  newValue?: string
+  status?: 'succeeded' | 'failed' | 'pending'
+  remark?: string
+}) => auditLogger.info('%j', data)
```
robertu7, line 82: For security reason, we should also log the IP address?

src/common/logger.ts
```diff
@@ -63,6 +63,24 @@ export const getLogger = (name: string) => {
   return newLogger
 }
 
+const auditLogger = getLogger('audit-log')
```
robertu7, line 66: So audit logs will be mix with other types of logs? What can we do for the performance of lambda-handler?

gary02: As audit log pushed to RDS is for analytic usage,      . the concern on lambda is not the speed but  whether it can finish processing 1 hour rotation log file in 15 mins.

gary02: RDS -> bigquery

robertu7: Shall we do a simple stress test?

src/common/logger.ts
```diff
@@ -15,11 +20,8 @@ export const contextStorage = new AsyncLocalStorage<
 >()
 
 const setContext = format((info, _) => {
-  const context = contextStorage.getStore()
-  if (context) {
-    info.requestId = context!.get('requestId')
-    // info.jobId = context!.get('jobId')
-  }
+  const context = contextStorage.getStore() || new Map()
+  info.requestId = context.get('requestId')
```
robertu7, line 24: https://github.com/thematters/matters-server/pull/3636/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80 

IP and UA?

gary02: here https://github.com/thematters/matters-server/pull/3636/files#diff-d6a621e5b8adb30e9f5b065e7d0c84185b704acb42df7009aa79d1fb397ecc40R88-R89

robertu7 (approved):
LGTM. And shall we add more types of audit logs (e.g. payment actions) as we mentioned in Notion doc even though  there is no need for product analysis now?

gary02:
> LGTM. And shall we add more types of audit logs (e.g. payment actions) as we mentioned in Notion doc even though there is no need for product analysis now?

sure,  I will add those audit logs in another PR

Issue #3372, opened 2023-05-25
Retire ethers.js


robertu7:
### Checklist
- [x] Replace ethers.js/eth-sig-util with viem
  - [x] `connector/blockchain/curation`
  - [x] `mutation/user/claimLogbook`
  - [x] `mutation/user/generateSigningMessage`
  - [x] `types/__test__/walletLogin.test`
  - [x] uses of `utils.isAddress` / `recoverPersonalSignature`
- [ ] ~~Store case sensitive ETH address instead of lowercased~~
- [ ] ~~Migrate old eth addresses~~

### Refs
* https://viem.sh/
* https://matterslab.slack.com/archives/C88CK7Q7L/p1684997544301659

byhow:
I can take a look at this issue if it has not been worked on yet, thanks

robertu7:
> I can take a look at this issue if it has not been worked on yet, thanks

reassigned to you, thanks!

PR #3638, opened 2023-10-07
migrate ether.js to viem

byhow:
Migrating all componenets and tests that are using ether.js to viem.sh

#3372


src/connectors/blockchain/curation.ts
```diff
@@ -49,86 +73,98 @@ const chainId = isProd
 // CurationContract
 
 export class CurationContract extends BaseContract {
```
byhow, line 75: Might take a while to migrate this particular contract logic to viem since it needs a major rewrite basically, apologize for the delay. Everything else should be good to go

robertu7: Thanks! We will take over this issue and keep building on top of your PR.

robertu7:
changed the base branch to `feat/viem` and ready to merge for the further development.

src/connectors/blockchain/curation.ts
byhow: @robertu7 Appreciate it! Sorry I was caught up with some family stuff over the past couple months. Glad that this PR got in :)

PR #3652, opened 2023-10-24
feat(myworks): api implement

gary02:


src/types/article.ts
```diff
@@ -160,6 +160,13 @@ export default /* GraphQL */ `
     "Total number of appreciations recieved of this article."
     appreciationsReceivedTotal: Int!
 
+
+    "Total number of donation recieved of this article."
+    donationCount: Int!
+
+    "Total number of readers of this article."
+    readerCount: Int!
```
robertu7, line 168: what's the cache TTL? or revalidate on `article_ga4_data` updates?

robertu7 (approved):

src/types/article.ts
gary02: right, here TTL should be set

gary02: right, here TTL should be set

PR #3708, opened 2023-11-24
feat(badges): Digital Nomad program

tx0c:
fix thematters/developer-resource#326

### DIFF
--- a/src/queries/user/badges.ts
+++ b/src/queries/user/badges.ts
@@ -1,4 +1,4 @@
-import { GQLUserInfoResolvers } from 'definitions'
+import { GQLBadgeType, GQLUserInfoResolvers } from 'definitions'
 
 const resolver: GQLUserInfoResolvers['badges'] = async (
   { id },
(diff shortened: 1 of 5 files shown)


src/queries/user/badges.ts
```diff
@@ -9,7 +9,14 @@ const resolver: GQLUserInfoResolvers['badges'] = async (
...
   }
 
-  return atomService.findMany({ table: 'user_badge', where: { userId: id } })
+  const badges = await atomService.findMany({
+    table: 'user_badge',
+    where: { userId: id },
+  })
+
+  return badges.map(({ type, extra }) => ({
+    type: `${type}${extra?.level || ''}` as GQLBadgeType,
```
robertu7, line 18: `extra` is arbitrary, how can it strong typed here?

robertu7 (approved):

src/queries/user/badges.ts
tx0c: we can do application level value clamp here; a typical use in DocumentDB

PR #3721, opened 2023-11-30
feat(badges): Digital Nomad program

tx0c:
for thematters/developer-resource#326 ;;

another entrance of set/reset badges; most badges should be given automatic but just in case OSS can override


robertu7 (approved):
better to revise the unit tests

PR #3735, opened 2023-12-21
feat(report): implement report API

gary02:


src/queries/system/report.ts
```diff
@@ -0,0 +1,32 @@
+import type { GQLReportResolvers, Draft } from 'definitions'
+
+import { NODE_TYPES } from 'common/enums'
+import { ServerError } from 'common/errors'
+import { toGlobalId } from 'common/utils'
+
+const report: GQLReportResolvers = {
+  id: ({ id }) => toGlobalId({ type: NODE_TYPES.Report, id }),
+  reporter: ({ userId }, _, { dataSources: { userService } }) =>
```
robertu7, line 9: shall we rename the `user_id` column to `reporter_id` so that `user_id` can be used for future use case like reporting a user?

robertu7 (approved):
LGTM

PR #3747, opened 2024-01-09
fix(with-oauth): protect some privacy data for user to view one's own…

tx0c:
… only


src/types/user.ts
```diff
@@ -404,7 +404,7 @@ export default /* GraphQL */ `
     articleCount: Int!
 
     "Number of comments posted by user."
-    commentCount: Int!
+    commentCount: Int! @auth(mode: "${AUTH_MODE.oauth}")
```
robertu7, line 407: this field is using by https://github.com/thematters/matters-web/blob/develop/src/components/ArticleDigest/Published/FooterActions/gql.ts#L18

robertu7 (approved):

Issue #3165, opened 2023-02-22
Backend filtering archived comments


devformatters:
### Description
Currently, we do comment filtering in client-side with rules [1],
* For level-1 comments: hide archived if it hasn't level-2 non-archived comments;
* For level-2 comments: hide archived;

However, user might need to click more than one time if comments of next pages are all filtered out...

### Tasks
- [x] Make sure API always returning the same size of result
- [ ] Return less info of archived comment

### References
1. https://github.com/thematters/matters-web/blob/98463d0ad5bb866ab3ddcf5b5435307d8bf1c30d/src/common/utils/comment.ts#L15
2. https://feedback.matters.town/bug-reports/p/010f7a
3. [Discussion](https://mattersnews.slack.com/archives/GA0R09FHN/p1676903814782049?thread_ts=1676878463.487499&cid=GA0R09FHN)
4. [Discussion](https://matterslab.slack.com/archives/GA0R09FHN/p1699606289565999)

PR #3753, opened 2024-01-11
feat(responses):  filter comments in backend

gary02:
- use `created_at` as cursor directly instead of `row_number`
- combine 2  db querying into one
- re https://github.com/thematters/matters-server/issues/3165
- refactor

### DIFF
--- a/src/connectors/articleService.ts
+++ b/src/connectors/articleService.ts
@@ -1310,7 +1310,6 @@ export class ArticleService extends BaseService {
   public findResponses = ({
     id,
     order = 'desc',
-    articleState = ARTICLE_STATE.active,
     after,
     before,
     first,
@@ -1320,7 +1319,6 @@ export class ArticleService extends BaseService {
   }: {
     id: string
     order?: string
-    articleState?: keyof typeof ARTICLE_STATE
     after?: { type: NODE_TYPES; id: string }
     before?: { type: NODE_TYPES; id: string }
     first?: number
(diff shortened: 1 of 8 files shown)


src/connectors/articleService.ts
```diff
@@ -1403,40 +1319,100 @@ export class ArticleService extends BaseService {
...
     first?: number
     includeAfter?: boolean
     includeBefore?: boolean
     articleOnly?: boolean
   }) => {
-    const query = this.makeResponseQuery({ id, order, state, articleOnly })
+    const subQuery = this.knexRO
+      .select(
+        this.knexRO.raw('COUNT(1) OVER() AS total_count'),
+        this.knexRO.raw('MIN(created_at) OVER() AS min_cursor'),
```
robertu7, line 1332: is there any issue since `created_at` might not be unique?

robertu7: and need to confirm if `created_at` is indexed.

robertu7 (approved):

src/connectors/articleService.ts
gary02: postgres datetime's  resolution is [1 microsecond](https://www.postgresql.org/docs/12/datatype-datetime.html) (1/1,000,000 second),  so the chance that items share exactly  same  `created_at` is extremely low

gary02: >and need to confirm if created_at is indexed.

both indexed in `comment`  and `article` table

gary02: > is there any issue since created_at might not be unique?
and `row_number` way (order by `created_at`)  have same issue when `created_at` not unique

PR #3757, opened 2024-01-12
feat(comment): hide most of info in archived/banned comments

gary02:


src/queries/comment/createdAt.ts
```diff
@@ -0,0 +1,21 @@
...
+  { viewer }
+) => {
+  const isActive = state === COMMENT_STATE.active
+  const isCollapsed = state === COMMENT_STATE.collapsed
+  const isAdmin = viewer.hasRole('admin')
+  if (isActive || isCollapsed || isAdmin) {
+    return createdAt
+  } else {
+    // invalid date
+    return new Date(0)
```
robertu7, line 17: should return `null` instead?

robertu7 (approved):

src/queries/comment/createdAt.ts
gary02: `createdAt: DateTime!`
can not return null here without update schema

PR #3771, opened 2024-01-20
feat: revise `Article.transactionsReceivedBy` to return users with anonymous

robertu7:
* update metadata to `blockchain_transaction` on fetch tx receipt;
* skip sending mail to sender if address is not matched;

### DIFF
--- a/src/connectors/blockchain/curation.ts
+++ b/src/connectors/blockchain/curation.ts
@@ -26,6 +26,9 @@ export interface CurationTxReceipt {
   events: CurationEvent[]
   txHash: string
   reverted: boolean
+  from: Address
+  to: Address | null
+  blockNumber: number
 }
 
 // constants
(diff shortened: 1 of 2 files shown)


src/types/article.ts
```diff
@@ -198,6 +197,9 @@ export default /* GraphQL */ `
     "Transactions history of this article."
     transactionsReceivedBy(input: TransactionsReceivedByArgs!): UserConnection! @complexity(multipliers: ["input.first"], value: 1)
 
+    "Donations of this article, grouped by sender"
+    donations(input: ConnectionArgs!): ArticleDonorConnection! @complexity(multipliers: ["input.first"], value: 1)
```
gary02, line 201: forget to rename `ArticleDonorConnection` to `ArticleDonationConnection`

gary02: so as `ArticleDonorEdge`,  `ArticleDonor`

gary02 (changes requested):

gary02 (approved):

PR #3786, opened 2024-01-30
feat: article versioning schema migration

gary02:
- database migrations for new article versioning design
- code base changes to adopt new article versioning design
- add table schema types
- improve BaseService and AtomService type check,   inject AtomService instance into BaseService as `this.models`
- remove aws, cloudfare from AtomService, as they are not db related
- remove dead code: sns related, ipfs sqs related, ipfs queue and others 
- refactors: move dataloaders from *Service to AtomService, move some bussiness logic from resolvers to service classes


src/mutations/article/updateArticleState.ts
```diff
@@ -34,12 +30,7 @@ const resolver: GQLMutationResolvers['updateArticleState'] = async (
       recipientId: user.id,
     })
   }
-
-  const { userName } = user
-  publicationQueue.refreshIPNSFeed({ userName })
```
robertu7, line 39: making article from active to inactive is expected to trigger a rebuild of author's IPNS feed, any reason to remove it here?

src/queries/article/revisedAt.ts
```diff
@@ -1,29 +1,15 @@
...
-
-  if (drafts.length <= 1) {
-    return
-  }
+  const articleVersion = await articleService.loadLatestArticleVersion(
+    articleId
+  )
 
-  return drafts[0].createdAt
+  return articleVersion.createdAt
```
robertu7, line 12: should return nullish if there is no revision.

src/queries/response/article/responses.ts
```diff
@@ -60,13 +58,13 @@ const resolver: GQLArticleResolvers['responses'] = async (
   )
 
   // re-process edges
-  const isDraft = (item: Draft | Comment): item is Draft => 'title' in item
+  const isArticle = (item: Article | Comment): item is Article =>
+    'title' in item
```
robertu7, line 62: `title` column will be removed from `article` table

src/queries/user/recommendation/readTagsArticles.ts
```diff
@@ -11,15 +9,15 @@ import {
...
+  { dataSources: { atomService } }
 ) => {
   if (!userId) {
     return connectionFromArray([], input)
   }
 
   const { take, skip } = fromConnectionArgs(input)
 
-  const [totalCount, tagDrafts] = await Promise.all([
+  const [totalCount, tagArticle] = await Promise.all([
```
robertu7, line 20: typo: "tagArticles"

src/mutations/article/updateArticleState.ts
gary02: `refreshIPNSFeed` method now only create ipns key in db and do not actually refresh IPNS, see https://github.com/thematters/matters-server/blob/master/src/connectors/articleService.ts#L315-L341 cc @tx0c

db/migrations/20240126160530_migrate_article_versioning.js
```diff
@@ -0,0 +1,209 @@
...
+    t.string('summary').notNullable()
+    t.boolean('summary_customized').notNullable()
+    t.bigInteger('content_id').unsigned().notNullable()
+    t.bigInteger('content_md_id').unsigned()
+    t.specificType('tags', 'text ARRAY').notNullable()
+    t.specificType('connections', 'text ARRAY').notNullable()
+    t.integer('word_count').notNullable()
+    t.string('data_hash')
+    t.string('media_hash')
+    t.enu('pin_state', ['pinned', 'pinning', 'failed']).notNullable()
```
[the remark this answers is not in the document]
gary02: resolved

src/definitions/chapter.d.ts
```diff
@@ -0,0 +1,9 @@
+export interface Chapter {
```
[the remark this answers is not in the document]
gary02: so as "Topic" ?

db/migrations/20240208144112_create_article_version_newest_view.js
robertu7, line 1: if our use case is to query the latest version of articles by given ids, shall we just use regular SQL instead of view to avoid the overhead of creating and querying a view?

robertu7 (approved):
💪

PR #3798, opened 2024-02-19
feat: article versioning api

gary02:


src/connectors/articleService.ts
```diff
@@ -334,14 +335,69 @@ export class ArticleService extends BaseService<Article> {
...
+          .where({ articleId })
+          .modify((builder) => {
+            if (onlyContentChange) {
+              builder.select(
+                '*',
+                this.knexRO.raw(
+                  'LAG(content_id, 1) OVER(order by id) AS last_content_id'
+                ),
+                this.knexRO.raw(
+                  'LAG(content_md_id, 1) OVER(order by id) AS last_content_md_id'
```
tx0c, line 365: the `order by id` is default `ASC` order, is it actually `first_content_md_id`?

the `ORDER BY id DESC` would be `last_content_md_id`

src/connectors/articleService.ts
```diff
@@ -334,14 +335,69 @@ export class ArticleService extends BaseService<Article> {
...
+            .orWhere(
+              'content_md_id',
+              '!=',
+              this.knexRO.ref('last_content_md_id')
+            )
+            .orWhere((whereBuilder) => {
+              // first version
+              whereBuilder
+                .where('last_content_id', null)
+                .where('last_content_md_id', null)
```
tx0c, line 388: these would convert to `= null` test, may not work as you may need a `IS NULL` test,

better use `.whereNull` ?
https://knexjs.org/guide/query-builder.html#wherenull

src/connectors/articleService.ts
gary02: I think here is right, as `LAG` returns value evaluated at the row that is offset rows **before** the current row within the partition ,  see https://www.postgresql.org/docs/8.4/functions-window.html

src/types/article.ts
```diff
@@ -239,13 +239,42 @@ export default /* GraphQL */ `
...
+     pageInfo: PageInfo!
+     edges: [ArticleVersionEdge]!
+  }
+
+  type ArticleVersionEdge {
+     node: ArticleVersion!
+     cursor: String!
+  }
+
+  type ArticleVersion implements Node {
```
robertu7, line 268: need to revise `node` and `nodes` resolvers to support `ArticleVersion`

src/types/__test__/1/article.test.ts
robertu7, line 1: add test: query article version by id

src/types/article.ts
gary02: [here](https://github.com/thematters/matters-server/pull/3798/files#diff-de19d85571b95ec048a6d56e10889ee21253a457e249ac35f81589238ed2b903R14) is

src/types/__test__/1/article.test.ts
gary02: https://github.com/thematters/matters-server/pull/3798/files/86b1b006de5ea05564a30339d5729399ddaeb684#diff-b27ea43616375e24ecfd4905af6c236f1e58dd077ce75fd00b353ea9ad68e958R306-R317 ?

src/connectors/articleService.ts
robertu7: maybe rename to `prev_content_id` for `LAG` and `next_content_id` for `LEAD`?

robertu7 (approved):

PR #3799, opened 2024-02-20
fix: avoid modifing object returned by dataloader

gary02:
dataloader return same objects in memory for same identifiers:

```javascript
const object =  await dataloader.load(1)   // object {id: 1,   content: 'test'}

delete object.id   // object now  {content: 'test'}

// code in other files
const object1 = await  dataloader.load(1)  // object1  {content: 'test'}


```


robertu7 (approved):
In JS practices, we try to avoid using `delete` 

FYI, https://javascript.plainenglish.io/avoid-the-delete-keyword-in-javascript-87ff2a47f26c

PR #3806, opened 2024-02-26
fix(comments): filter archived / banned comments by default

gary02:
filter archived/banned comments except those which have active/collapsed sub comments.


robertu7 (approved):
LGTM

PR #3818, opened 2024-03-04
feat: update article editon logic

gary02:


src/types/__test__/1/editArticle.test.ts
```diff
@@ -0,0 +1,1046 @@
...
+        input: {
+          id: articleGlobalId,
+          title,
+        },
+      },
+    })
+    expect(sameData.editArticle.title).toBe(title)
+    expect(sameData.editArticle.revisionCount).toBe(1)
+
+    // empty string is allowed
```
robertu7, line 234: same comment as above cuz [title is required](https://github.com/thematters/matters-server/blob/develop/src/mutations/article/publishArticle.ts#L45-L47), pls confirm

robertu7 (approved):

PR #3849, opened 2024-03-28
fix(related-articles): fix relatedArticle resolver that return only 1 article in some case

gary02:
objects returned by `findByAuthor` should contain `id`,  as inside the helper function `addRec`,  `_.uniqBy(..., 'id')` is called, passing objects w/o `id` field will filter and return only one object, which is the cause some articles' return only 1 related article.


gary02:
Objects returned by findByAuthor not containing `id` is also the root cause that `addRec` fail to filter the parent article out.

tx0c (approved):

PR #3852, opened 2024-04-01
fix(sendmail): check users' email before sendmail

gary02:


src/connectors/paymentService.ts
```diff
@@ -1088,7 +1100,7 @@ export class PaymentService extends BaseService<Transaction> {
     }
 
     // send email to sender
-    if (sender) {
+    if (sender && sender.email) {
```
tx0c, line 1103: if (sender?.email) {

tx0c (approved):

PR #3909, opened 2024-05-14
feat(draft): temporarily use article version as draft in article drafts API

gary02:
matters-web 4.30.2 (prod) still relies on those APIs for article revision, but server does not create new drafts but article-versions, we temporarily fake the APIs to fix the problem


robertu7 (approved):
Let's try

PR #3917, opened 2024-05-16
fix(atomservice): work around knexSnakeCaseMappers edge case

gary02:
`cardLast4` is transformed to `card_last4` instead of `card_last_4`


robertu7 (approved):
nice catch...

Issue #3953, opened 2024-05-28
Implement APIs for Editor / Detail Page


gary02:
## Task

- [x] implement API design in #3952 
- [x] revise `submitReport` api
- [x] revise `node` api
- [x] revise `directImageUpload` api 
- [x] test `node` api
- [x] `putJournal` api ratelimit
- [x] revise `putComment` api

## Story

- https://github.com/thematters/developer-resource/issues/436
- https://github.com/thematters/developer-resource/issues/437

PR #3980, opened 2024-06-07
feat(moment): impl moment api

gary02:
resolve https://github.com/thematters/matters-server/issues/3953


src/connectors/journalService.ts
```diff
@@ -0,0 +1,144 @@
...
+    const content = sanitizeHTML(data.content)
+    const journal = await this.models.create({
+      table: 'journal',
+      data: {
+        authorId: user.id,
+        content: content,
+        state: JOURNAL_STATE.active,
+      },
+    })
+    if (data.assetIds && data.assetIds.length > 0) {
```
robertu7, line 45: check asset author and type also?

src/types/journal.ts
```diff
@@ -0,0 +1,51 @@
+import { AUTH_MODE, NODE_TYPES } from 'common/enums'
+
+export default /* GraphQL */ `
+  extend type Mutation {
+    putJournal(input: PutJournalInput!): Journal! @auth(mode: "${AUTH_MODE.oauth}") @rateLimit(limit:2, period:300)
```
robertu7, line 5: confirm with product or CC on rate limiting?

src/queries/journal/comments.ts
```diff
@@ -0,0 +1,79 @@
...
+  })
+
+  const where: CommentFilter = {
+    type: COMMENT_TYPE.journal,
+    targetId: id,
+    targetTypeId,
+  }
+  if (rest.filter) {
+    const { parentComment, author, state } = rest.filter
+    if (parentComment || parentComment === null) {
```
robertu7, line 48: can ignore `parentComment` for journal comments?

robertu7 (approved):

src/connectors/cloudflare/index.ts
```diff
@@ -65,7 +64,11 @@ export class CloudflareService {
     return key
   }
 
-  baseUploadFile = async (folder: GQLAssetType, upload: any, uuid: string) => {
+  public baseUploadFile = async (
+    folder: GQLAssetType,
+    upload: any,
```
byhow, line 69: not a big fan whenever it comes to `any`. We can explicitly type it with just the types we will be using here:
`const { createReadStream, mimetype, filename } = upload`

src/connectors/journalService.ts
```diff
@@ -0,0 +1,144 @@
...
+            data: { assetId, journalId: journal.id },
+          })
+        )
+      )
+    }
+    return journal
+  }
+
+  public delete = async (id: string, user: User) => {
+    if (![USER_STATE.active, USER_STATE.banned].includes(user.state as any)) {
```
byhow, line 59: ditto on using `any`

src/connectors/journalService.ts
```diff
@@ -0,0 +1,144 @@
...
+      targetId: user.id,
+    })
+    if (isBlocked) {
+      throw new ForbiddenError(`user ${id} is blocked by target author`)
+    }
+    return this.models.upsert({
+      table: 'action_journal',
+      where: { targetId: id, userId: user.id },
+      create: { targetId: id, userId: user.id, action: 'like' },
+      update: {},
```
byhow, line 108: can we omit this update : {}?

src/connectors/journalService.ts
gary02: so as content length

src/types/journal.ts
gary02: just a temp value by now,  will  be update in future PR

PR #3998, opened 2024-06-17
feat(journal): add user writings API

gary02:


src/connectors/userWorkService.ts
```diff
@@ -0,0 +1,102 @@
...
+      )
+      .as('t2')
+
+    const query = knexRO
+      .from(subQuery)
+      .orderBy('created_at', 'desc')
+      .limit(take)
+
+    if (after) {
+      const cursor = knexRO(after.type.toLowerCase())
```
robertu7, line 80: no need to `await`?

robertu7 (approved):

src/connectors/userWorkService.ts
gary02: This is a subquery.

PR #4021, opened 2024-06-30
feat(moment): add short_hash field

gary02:


src/connectors/momentService.ts
```diff
@@ -63,6 +64,7 @@ export class MomentService {
     const moment = await this.models.create({
       table: 'moment',
       data: {
+        shortHash: nanoid(),
```
robertu7, line 67: we should make it a explicitly function call by providing necessary params.

robertu7: also for article shortHash

robertu7 (approved):

PR #4026, opened 2024-07-08
feat(campaign): add CRUD APIs

gary02:


src/common/utils/validator.ts
```diff
@@ -129,3 +129,10 @@ export const isValidTransactionHash = (hash: string): boolean => {
 
   return /^0x([a-fA-F0-9]{64})$/.test(hash)
 }
+
+export const isValidUrl = (url: string): boolean => {
```
robertu7, line 133: can align with [client-side](https://github.com/thematters/matters-web/blob/develop/src/common/utils/url.ts#L1)

src/connectors/campaignService.ts
```diff
@@ -0,0 +1,209 @@
...
+        writingPeriod: toDatetimeRangeString(
+          writingPeriod[0],
+          writingPeriod[1]
+        ),
+        state: state || CAMPAIGN_STATE.pending,
+        creatorId,
+      },
+    })
+
+  public updateStages = async (campaignId: string, stages: Stage[]) => {
```
robertu7, line 81: use `stages` as new stages would be more simple than update the diff?

src/queries/campaign/index.ts
```diff
@@ -0,0 +1,60 @@
...
+import description from './description'
+import name from './name'
+import participants from './participants'
+import stageName from './stage/name'
+import stages from './stages'
+
+const schema: GQLResolvers = {
+  Query: {
+    campaign: (_, { input: { shortHash } }, { dataSources: { atomService } }) =>
+      atomService.findUnique({
```
robertu7, line 17: exclude inactive campaign for non-admin users?

src/queries/campaign/stage/name.ts
```diff
@@ -0,0 +1,17 @@
+import type { GQLCampaignStageResolvers } from 'definitions'
+
+const resolver: GQLCampaignStageResolvers['name'] = async (
+  { id, name },
+  _,
+  { viewer, dataSources: { translationService } }
+) => {
+  const translation = await translationService.findTranslation({
+    table: 'campaign',
```
robertu7, line 9: campaign_stage

src/types/__test__/2/campaign.test.ts
```diff
@@ -0,0 +1,436 @@
...
+    start: new Date('2024-01-01'),
+    end: new Date('2024-01-02'),
+  }
+  const writingPeriod = {
+    start: new Date('2024-01-03'),
+    end: new Date('2024-01-04'),
+  }
+  const stages = [
+    {
+      name: translations,
```
robertu7, line 86: should use different translations as stage name 

ref: comment `src/queries/campaign/stage/name.ts#L9`

robertu7 (changes requested):

src/connectors/campaignService.ts
gary02: new stages will have new Ids and thus create new translation records ?

src/queries/campaign/stages.ts
```diff
@@ -0,0 +1,10 @@
+import type { GQLWritingChallengeResolvers } from 'definitions'
+
+const resolver: GQLWritingChallengeResolvers['stages'] = async (
+  { id },
+  _,
+  { dataSources: { atomService } }
+) =>
+  atomService.findMany({ table: 'campaign_stage', where: { campaignId: id } })
```
robertu7, line 8: return in order?

src/connectors/campaignService.ts
robertu7: yeah, and it's more easy to [return in order](https://github.com/thematters/matters-server/pull/4026/files#r1667990887)?

robertu7 (approved):

PR #4040, opened 2024-07-09
feat(campaign): add `User.campaigns` API

gary02:


src/types/campaign.ts
```diff
@@ -119,7 +119,7 @@ export default /* GraphQL */ `
   }
 
   input CampaignArticlesFilter{
-    stage: String!
+    stage: ID!
```
robertu7, line 122: Beware that `Campaign.articles` is not implemented yet

robertu7 (approved):

PR #4105, opened 2024-08-02
feat(upload): add ratelimit and auditlog

gary02:


src/types/system.ts
```diff
@@ -1,5 +1,7 @@
 import { AUTH_MODE, CACHE_TTL, NODE_TYPES, SCOPE_GROUP } from 'common/enums'
 
+const UPLOAD_RATE_LIMIT = 40 // 20 pictures per 12 minutes, `directImageUpload` called twice for each picture
```
robertu7, line 3: 20？

robertu7 (approved):

src/types/system.ts
gary02: as stated in comment, `directImageUpload` is called twice for each picture

PR #4117, opened 2024-08-08
feat(spam-detection): detect spam and exclude spam from article feeds

gary02:


src/connectors/spamDetector.ts
```diff
@@ -0,0 +1,36 @@
...
+      url: this.apiUrl,
+      data: {
+        text,
+      },
+    }
+
+    try {
+      const response = await axios(config)
+      return response.data.score
+    } catch (error) {
```
robertu7, line 31: need to add an CloudWatch alarm to monitor the API error rate?

src/connectors/systemService.ts
```diff
@@ -150,6 +152,21 @@ export class SystemService extends BaseService<BaseDBSchema> {
     return false
   }
 
+  /**
+   * Get the spam threshold from `feature_flag` table
+   * Use to determine whether a article is spam by its spam score
+   */
+  public getSpamThreshold = async (): Promise<number | null> => {
```
robertu7, line 159: cache result to avoid repetitive query for a period of time?

db/migrations/20240806102534_alter_article_spam.js
```diff
@@ -0,0 +1,21 @@
+const table = 'article'
+
+exports.up = async (knex) => {
+  await knex.schema.table(table, (t) => {
+    t.float('spam_score').nullable()
+    t.boolean('is_spam').nullable()
+
+    t.index('spam_score')
```
robertu7, line 8: do we need a timestamp to record the time that run the spam check, if any further usage?

robertu7 (approved):

db/migrations/20240806102534_alter_article_spam.js
gary02: I log the check time in lambda logging by now
https://github.com/thematters/spam-detection-serverless/blob/main/app/app.py#L29-L31

gary02: add more logging https://github.com/thematters/matters-server/pull/4117/commits/744abf95f16113d712244c85bb83c3bc76b16f3f

Issue #3971, opened 2024-06-04
Edge cases for article translation


robertu7:

lizhineng:
I forked the project yesterday and am still familiarizing myself with the repository. Please correct me if I'm wrong.

- Does the bug indicate that we need to handle situations when the translation API runs out of quota?
- Should we replace Google Translate with another provider due to pricing concerns, or are we supposed to support multiple translation providers with load balancing behind the scenes? This would help us avoid relying on a single translation API, reducing the risk of a single point of failure and improving availability.

Additionally, I noticed that the translation feature also translates _@mention_. I think we should leave it as is—what are your thoughts?

PR #4155, opened 2024-08-27
Strip @mention for article language resolver

lizhineng:
This is part of the tasks assigned in #3971, the article content would first be stripped of all mentions before being extracted for the excerpt. This is my first PR and I hope I didn't miss anything.


gary02 (approved):
LGTM

robertu7 (approved):
@lizhineng thank you for your contribution! Really appreciate you taking time to fix this issue! 

FYI, the translation feature is not good enough and buggy as the demand increases, #3971 only with some easy fixes. We are supposed to refactor it with better and cheaper solution (e.g. language model, Microsoft Translator), free feel to join the discussion and left your thoughts in #3971.

PR #4164, opened 2024-09-02
feat(block): update mutation logic for blocked users

gary02:


src/mutations/article/appreciateArticle.ts
robertu7, line 1: As the PRD, blockee cannot appreciate the blocker's article.

robertu7 (dismissed):

src/mutations/article/appreciateArticle.ts
gary02: Right, I mix up appreciate and donation

PR #4319, opened 2025-03-04
feat(cache): introduce separate object cache redis configuration

gitwoz:
This change adds support for a separate Redis configuration for object caching, allowing more granular control over caching strategies. Key modifications include:
- Added new environment variables for object cache host and port
- Updated connections and services to use a dedicated object cache Redis instance
- Refactored references to use `objectCacheRedis` across various services and resolvers


ham-76 (approved):
LGTM

